PR #8, opened 2020-12-02
Update dependencies

andrewrlee:
Also move to use declaration overrides for describing additional possible fields in session data

Also removing/moving unnecessary prod dependencies

Moving jest tests to run in parallel


package.json
```diff
@@ -17,9 +17,9 @@
     "watch-node-feature": "export $(cat feature.env) && nodemon --watch dist/ $NODE_DEBUG_OPTION dist/server.js | bunyan -o short",
     "start-feature:dev": "npm run build && concurrently -k -p \"[{name}]\" -n \"Views,TypeScript,Node,Sass\" -c \"yellow.bold,cyan.bold,green.bold,blue.bold\" \"npm run watch-views\" \"npm run watch-ts\" \"npm run watch-node-feature\" \"npm run watch-sass\"",
     "record-build-info": "node ./bin/record-build-info",
-    "lint": "eslint . --cache --max-warnings 200",
+    "lint": "eslint . --cache",
```
andrewrlee, line 20: no need for allowing warnings - we currently don't have any 🏅

package.json
```diff
@@ -17,9 +17,9 @@
...
     "typecheck": "tsc",
-    "test": "jest --runInBand",
+    "test": "jest",
```
andrewrlee, line 22: run tests in parallel by default

package.json
```diff
@@ -87,9 +87,8 @@
     ]
   },
   "dependencies": {
-    "@types/connect-flash": "0.0.35",
```
andrewrlee, line 90: should be a dev dependency

package.json
```diff
@@ -98,18 +97,15 @@
     "connect-redis": "^5.0.0",
     "cookie-session": "^1.4.0",
     "csurf": "^1.11.0",
-    "date-fns": "^2.16.1",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
     "express-request-id": "^1.4.1",
     "express-session": "^1.17.1",
-    "govuk-elements-sass": "^3.1.3",
```
andrewrlee, line 106: this, `joi` and `date-fns` not used

server/@types/express/index.d.ts
```diff
@@ -0,0 +1,9 @@
+export default {}
+
+declare module 'express-session' {
+  // Declare that the session will potentially contain these additional fields
+  interface SessionData {
+    returnTo: string
+    nowInMinutes: number
+  }
+}
```
andrewrlee, line 9: Really fiddly but added `./server/@types` as a definition root which allows overriding types in other packages.

Here we want to be able to say that the session might contain `returnTo` and `nowInMinutes`

We probably will also want to add the user object to res.locals but not currently required

server/authentication/auth.ts
```diff
@@ -28,16 +28,6 @@ const authenticationMiddleware: AuthenticationMiddleware = verifyToken => {
   }
 }
 
-export interface UserDetails {
-  username: string
-  token: string
-  authSource: string
-}
-
-export interface UserRequest extends Request {
```
andrewrlee, line 37: these weren't used

server/data/tokenVerification.ts
```diff
@@ -15,15 +15,7 @@ function getApiClientToken(token: string) {
     })
 }
 
-export interface VerifiableRequest extends Request {
```
andrewrlee, line 18: this is provided in our express type overrides

server/middleware/authorisationMiddleware.ts
```diff
@@ -6,7 +6,7 @@ import logger from '../../log'
 export default function authorisationMiddleware(authorisedRoles: string[] = []): RequestHandler {
   return (req, res, next) => {
     if (res.locals && res.locals.user && res.locals.user.token) {
-      const { authorities: roles = [] } = jwtDecode(res.locals.user.token)
+      const { authorities: roles = [] } = jwtDecode(res.locals.user.token) as { authorities?: string[] }
```
andrewrlee, line 9: `jwtDecode` returns unknown - overriding this to say that whatever it returns may or may not contain an array of strings

tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "typeRoots": ["./server/@types", "./node_modules/@types"]
```
andrewrlee, line 18: adding our custom type overrides

package.json
psoleckimoj: We have ours set to zero, not sure what difference that will make if we have any here?

package.json
```diff
@@ -118,38 +114,39 @@
...
-    "jest": "^26.5.3",
+    "jest": "^26.6.3",
     "jest-html-reporter": "^3.3.0",
-    "jest-junit": "^11.1.0",
+    "jest-junit": "^12.0.0",
     "jsonwebtoken": "^8.5.1",
-    "lint-staged": "^10.4.1",
-    "mocha": "^8.1.3",
+    "lint-staged": "^10.5.2",
     "mocha-junit-reporter": "^2.0.0",
```
psoleckimoj, line 143: Is there any point using this if we're don't have mocha any more or is it used for something else?

package.json
andrewrlee: Ah, good spot - I saw it defaulted to -1 which I assume would be the same a 0 but in reality disables the failing.

I've explicitly set this to 0 and now it fails with any warnings

package.json
andrewrlee: Based on this cypress config:

```
{
  "reporterEnabled": "spec, mocha-junit-reporter",
  "reporterOptions": {
    "mochaFile": "test_results/cypress/results-[hash].xml"
  }
}
```

this is actually used by cypress to generate junit test result files in this location:

`test_results/cypress/results-some-hash.xml`

These are then uploaded to cypress through this config: 
```
      - store_test_results:
          path: test_results
```

The results can then be seen in the circle ci ui:
<kbd>
Screenshot 2020-12-07 at 08 29 55
</kbd>

(Fixed the file path so this actually works!)

brightonsbox (approved):
LGTM!

PR #9, opened 2020-12-14
Reducing docker image size

andrewrlee:
Also
* Update dependencies
* Adopt standard github gitignore for node projects
* Using body typography


Dockerfile
```diff
@@ -43,19 +47,20 @@ RUN apt-get autoremove -y && \
...
-RUN npm prune --production
+COPY --from=build --chown=appuser:appgroup \
+        /app/node_modules ./node_modules
 
 EXPOSE 3000
+ENV NODE_ENV='production'
 USER 2000
 
-CMD [ "npm", "start" ]
+CMD [ "npm", "start" ]
```
psoleckimoj, line 66: Needs a newline 😄

psoleckimoj (approved):
Other than needing that newline, LGTM

Dockerfile
andrewrlee: gah.. will add in a follow up

PR #28, opened 2021-05-24
Moving to group middleware into related modules

andrewrlee:
This follows the pattern established in prison staff hub.

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -10,12 +10,14 @@ declare module 'express-session' {
 
 export declare global {
   namespace Express {
+    interface User {
+      username: string
+      token: string
+      authSource: string
+    }
+
     interface Request {
       verified?: boolean
-      user: {
-        username: string
-        token: string
-      }
     }
   }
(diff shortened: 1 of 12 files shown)


server/data/tokenVerification.test.ts
```diff
@@ -7,6 +7,7 @@ describe('token verification api tests', () => {
   let fakeApi: nock.Scope
 
   beforeEach(() => {
+    config.apis.tokenVerification.url = 'http://localhost:8100'
```
andrewrlee, line 10: fixing this test as it fails when a local .env file is present - should probably move to just injecting the url

server/@types/express/index.d.ts
```diff
@@ -10,12 +10,14 @@ declare module 'express-session' {
 
 export declare global {
   namespace Express {
+    interface User {
```
andrewrlee, line 13: this overrides the Express.User which is used everywhere rather than just the user object attached to the request

lbennett-moj (approved):

PR #32, opened 2021-06-16
Fixing configuration of security audit

brightonsbox:
The "medium" configuration was not an acceptable term, should be one of "info", "low", "moderate", "high" or "critical". This meant that it would not pick up on anything!


psoleckimoj (approved):
LGTM

PR #33, opened 2021-06-24
Alee/update dependencies

andrewrlee:


.circleci/config.yml
```diff
@@ -1,26 +1,28 @@
...
+parameters:
+  alerts-slack-channel:
+    type: string
+    default: hmpps_typescript_notifications
+
 executors:
+  builder:
+    docker:
+      - image: 'cimg/node:14.17-browsers'
+    working_directory: ~/app
```
andrewrlee, line 16: move to not use the hmpps/node executor as it means you have to specify version in lots of places and it's not obvious which version of node you are using and theoretically it should match the int-tests and Docker file version

.circleci/config.yml
```diff
@@ -177,9 +175,11 @@ workflows:
           context:
             - hmpps-common-vars
       - hmpps/npm_security_audit:
+          slack_channel: << pipeline.parameters.alerts-slack-channel >>
           context:
             - hmpps-common-vars
       - hmpps/veracode_policy_scan:
+          slack_channel: << pipeline.parameters.alerts-slack-channel >>
```
andrewrlee, line 182: Moved these to use shared param so its more obvious where the value is coming from rather than using defaults

.circleci/config.yml
```diff
@@ -1,26 +1,28 @@
...
+
 executors:
+  builder:
+    docker:
+      - image: 'cimg/node:14.17-browsers'
+    working_directory: ~/app
   integration-tests:
     docker:
-      - image: cimg/node:14.15-browsers
+      - image: cimg/node:14.17-browsers
```
petergphillips, line 19: noticed we've skipped over `-16` here.  Is that an argument for saying that it should actually be `14-browser` and let it float?

.circleci/config.yml
petergphillips: agreed, sounds a bit naff to have to specify in multiple places but wondering if there is a better way :thinking:

petergphillips: if you declared `14.17-browsers` as a parameter like `alerts-slack-channel` then you could just reference the parameter in all usages in the file and then wouldn't have the issue?

petergphillips (approved):
LGTM

petergphillips (approved):
LGTM

.circleci/config.yml
andrewrlee: I did that originally and undid it 😄  

Will try and tidy up in a follow up

.circleci/config.yml
andrewrlee: as discussed - there isn't a 14-browser as circle recommend pinning to a specific version

PR #34, opened 2021-06-28
Use node executor and parameterise the version

markreesmoj:


.circleci/config.yml
```diff
@@ -9,11 +9,11 @@ parameters:
...
+    default: 14.17-browsers
+
 executors:
-  builder:
-    docker:
-      - image: 'cimg/node:14.17-browsers'
-    working_directory: ~/app
   integration-tests:
     docker:
       - image: cimg/node:14.17-browsers
```
andrewrlee, line 19: can use the parameter here too

petergphillips, line 19: would also be good to reference `node-version` here

petergphillips (dismissed):
LGTM

.circleci/config.yml
```diff
@@ -9,20 +9,22 @@ parameters:
...
+    type: string
+    default: 14.17-browsers
+
 executors:
-  builder:
-    docker:
-      - image: 'cimg/node:14.17-browsers'
-    working_directory: ~/app
   integration-tests:
     docker:
```
petergphillips, line 18: think you can now change this to use the `node_redis` executor from hmpps too :grin:

markreesmoj: Ha I saw that PR coming and that's why I didn't bother to parameterise this section in the first place! One more push

andrewrlee (approved):
LGTM

.circleci/config.yml
```diff
@@ -92,7 +91,10 @@ jobs:
           path: test-results/unit-test-reports.html
 
   integration_test:
-    executor: integration-tests
+    executor:
+      name: hmpps/node_redis
+      node_tag: << pipeline.parameters.node-version >>
+      redis_tag: buster
```
petergphillips, line 97: think this should probably be `4-buster` to match the elasticache image, but was already set incorrectly so :shrug:

PR #39, opened 2021-07-01
Updating dependencies and improving README

andrewrlee:


.circleci/config.yml
```diff
@@ -164,6 +168,33 @@ workflows:
             - integration_test
             - build_docker
 
+#      - request-preprod-approval:
```
andrewrlee, line 171: Wouldn't usually commit commented out code but following the kotlin template project example. 😄

markreesmoj (approved):
👍

PR #41, opened 2021-07-26
Moving cypress tests to typescript

andrewrlee:


server/data/restClient.ts
```diff
@@ -102,11 +98,7 @@ export default class RestClient {
     }
   }
 
-  async stream({
-    path = null,
-    headers = {},
-    errorLogger = this.defaultErrorLogger,
```
andrewrlee, line 108: @jsrobertson noticed that this was broken - the direct reference to the function meant that `this` got removed which lead to an undefined access error on line 48

yasinmustafa (approved):
LGTM

PR #42, opened 2021-08-06
DT-2404: 💄 Switch to sign in / out instead of login / out

petergphillips:


docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
       test: ["CMD", "curl", "-f", "http://localhost:8080/auth/health"]
     environment:
       - SPRING_PROFILES_ACTIVE=dev
-      - DELIUS_ENABLED=false
```
petergphillips, line 25: `dev` profile disables delius anyway

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
...
+      - APPLICATION_AUTHENTICATION_UI_ALLOWLIST=0.0.0.0/0
```
petergphillips, line 25: allows access to the client configuration within auth

.husky/.gitignore
```diff
@@ -1 +0,0 @@
-_
```
petergphillips, line 1: removed as not needed in latest version of husky

andrewrlee (approved):
LGTM

PR #46, opened 2021-08-23
Remove unused connect-flash

psoleckimoj:


.circleci/config.yml
```diff
@@ -101,6 +101,8 @@ jobs:
       redis_tag: buster
     steps:
       - checkout
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package-lock.json" }}
```
andrewrlee, line 105: doesn't the attach workspace on line 106 also attach the node_modules that was persisted on line 50?
Is this needed?

If it is needed, does this mean we can drop the restore cache on line 125?

server/middleware/setUpAuthentication.ts
```diff
@@ -12,7 +11,6 @@ export default function setUpAuth(): Router {
 
   router.use(passport.initialize())
   router.use(passport.session())
-  router.use(flash())
```
andrewrlee, line 15: Fine for this to go now, but flash is really useful when using post/redirect/get which would generally recommend services to use rather than write their own session management abstraction to handle this use case.

We might add a form example using classvalidator in which case this might be re-added

PR #50, opened 2021-09-07
Fix docker build failing, and reduce image size

mattops:


Dockerfile
```diff
@@ -15,12 +15,18 @@ RUN addgroup --gid 2000 --system appgroup && \
...
-    apt-get upgrade -y
+    apt-get upgrade -y && \
+    apt-get autoremove -y && \
+    rm -rf /var/lib/apt/lists/*
 
 # Stage: build assets
 FROM base as build
 
-RUN apt-get install -y make python g++
+ARG BUILD_NUMBER=1_0_0
```
petergphillips, line 25: I thought we decided we didn't need these two lines as they were in the base?

Dockerfile
```diff
@@ -15,12 +15,18 @@ RUN addgroup --gid 2000 --system appgroup && \
...
+ARG GIT_REF=not-available
+
+RUN apt-get update && \
```
petergphillips, line 28: how come you've got an `update` here but not  an `upgrade`?

Dockerfile
mattops: They are used at lines 37,38 - this is what was causing the build to fail.  It seams in a multistage build the ARG's are not available across stages.

Dockerfile
mattops: Because this is stage that is thrown away - essentially all we need is to pull in the package lists so we can install the build dependancies (make python g++).

andrewrlee (approved):
LGTM

Dockerfile
petergphillips: isn't the `update` superfluous in that we did an `update` in `base` anyway

petergphillips: why do we need to do an extra one here?

mattops: > isn't the `update` superfluous in that we did an `update` in `base` anyway

No, because we cleaned up the package lists with `rm -rf /var/lib/apt/lists/*` at line 20 - this is to save space in the final image (approx 20MB).  Previously this clean up command was in the final stage - however that was pointless in terms of saving space in the image.

The `update` here will again pull down the package lists (maybe wasting a bit of time) but we dont need to clean up because this stage is thrown away.

PR #51, opened 2021-09-28
Remove use of semi colons before arrays

garethmdavies:
remove use of `;]` to match rest of code base where eslint rules have semi-colon usage disabled


psoleckimoj (approved):
LGTM!

PR #62, opened 2021-11-05
NN-3747 fixing json structure for the stubUserRoles call and populati…

richardpopple:
…ng the user directly and not from the request because passport isn't in the test stack


petergphillips (approved):
LGTM

PR #63, opened 2021-11-10
Bumping node version

andrewrlee:
Also fixing open handle in test and bumping dependencies


server/services/healthCheck.test.ts
```diff
@@ -36,22 +36,22 @@ describe('Healthcheck', () => {
...
   return () =>
-    new Promise((resolve, _reject) =>
+    new Promise((resolve, _reject) => {
       resolve({
         name: `${name}`,
         status: 'ERROR',
         message: 'some error',
       })
-    )
+    })
```
andrewrlee, line 56: new [eslint rule](https://eslint.org/docs/rules/no-promise-executor-return)

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Stage: base image
-FROM node:14.17-buster-slim as base
+FROM node:16.13-bullseye-slim as base
```
andrewrlee, line 2: Floating at minor version to keep in line with the circle image

petergphillips (approved):
LGTM

PR #65, opened 2021-12-03
Update dependencies and move to NPM v8

psoleckimoj:
- Dependency updates (except redis as type defs not up-to-date)
- Move to npm 8 which is standard with node 16


andrewrlee (approved):
LGTM

PR #66, opened 2022-01-04
Upgrading dependencies

andrewrlee:


package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "prepare": "husky install",
     "copy-views": "cp -R server/views dist/server/",
-    "compile-sass": "sass --no-source-map --load-path=node_modules/govuk-frontend --load-path=node_modules/@ministryofjustice/frontend ./assets/sass/application.sass:./assets/stylesheets/application.css ./assets/sass/application-ie8.sass:./assets/stylesheets/application-ie8.css --style compressed",
+    "compile-sass": "sass --quiet-deps --no-source-map --load-path=node_modules/govuk-frontend --load-path=node_modules/@ministryofjustice/frontend --load-path=. ./assets/sass/application.sass:./assets/stylesheets/application.css ./assets/sass/application-ie8.sass:./assets/stylesheets/application-ie8.css --style compressed",
```
andrewrlee, line 10: reduce logging or errors related to [this](https://github.com/alphagov/govuk-frontend/issues/2238) issue

Also need to add new load path for `.` as govuk frontend now explicitly reference govuk node modules as a dependency related to this: https://github.com/ministryofjustice/moj-frontend/issues/147

server/middleware/setUpWebSession.ts
```diff
@@ -1,25 +1,20 @@
...
-
 export default function setUpWebSession(): Router {
+  const client = createRedisClient(true)
+  client.connect()
+
   const router = express.Router()
   router.use(
     session({
-      store: new RedisStore({ client }),
+      store: new RedisStore({ client: client as unknown as Client }),
```
andrewrlee, line 17: bit gnarly - waiting on redis store to update to new version of redis client: 
https://github.com/tj/connect-redis/pull/337

package.json
psoleckimoj: Do we need the second one? Or do we only import bits of govuk? ...and a separate issue, what point are we ditching IE8 support? I don't believe it's in our supported browsers.

andrewrlee: We need all 3 - or we have to start referring to things with full paths. 
The new load path is used only to bodge around the change in the moj-frontend and how they refer to assets from the gov uk frontend library. 

I think we still have users in probation who use IE8, we did the last time I checked..

psoleckimoj: Thanks for the explanation @andrewrlee 👍

psoleckimoj (approved):
LGTM!

PR #69, opened 2022-02-04
INC-163 Timeout Fix - Correctly sets the timeout for a HttpAgent

Mjwillis:


server/config.ts
```diff
@@ -15,11 +15,11 @@ function get<T>(name: string, fallback: T, options = { requireInProduction: fals
 const requiredInProduction = { requireInProduction: true }
 
 export class AgentConfig {
-  maxSockets: 100
+  timeout: number
 
-  maxFreeSockets: 10
-
-  freeSocketTimeout: 30000
+  constructor(timeout = 8000) {
```
andrewrlee, line 20: FYI think you can do, and then not define the field:
```constructor(readonly timeout = 8000) {}```

(You can definitely do it with `constructor(private readonly ...){}` - haven't done it with public fields)

andrewrlee (approved):
LGTM, thanks for fixing!

server/config.ts
Mjwillis: Agree about the class but didn't want to change too many files at this point

PR #70, opened 2022-02-14
⬆️ update dependencies and 💄add no-only-tests linting rule for cypress

connormaglynn:


package.json
```diff
@@ -86,19 +86,19 @@
...
     "bunyan": "^1.8.15",
     "bunyan-format": "^0.2.1",
     "compression": "^1.7.4",
     "connect-flash": "^0.1.1",
-    "connect-redis": "^6.0.0",
+    "connect-redis": "^6.1.1",
     "cookie-session": "^2.0.0",
     "csurf": "^1.11.0",
-    "dotenv": "^14.3.0",
+    "dotenv": "^16.0.0",
```
connormaglynn, line 98: only effects values starting with **#** or containing **`**. Don't believe we have either 👍

connormaglynn: changelog: https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md

package.json
```diff
@@ -86,19 +86,20 @@
...
     "agentkeepalive": "^4.2.0",
-    "applicationinsights": "^2.2.0",
+    "applicationinsights": "^2.2.1",
     "body-parser": "^1.19.1",
     "bunyan": "^1.8.15",
     "bunyan-format": "^0.2.1",
     "compression": "^1.7.4",
     "connect-flash": "^0.1.1",
-    "connect-redis": "^6.0.0",
+    "connect-redis": "^6.1.1",
```
andrewrlee, line 95: Wonder if this means we can tidy up the redis stuff a bit - I think it means you can use redis client with legacy mode.. will have to have a look later

PR #74, opened 2022-03-18
Fixing docker caching issue

andrewrlee:
Need to refer to build args before calling apt-get upgrade otherwise the set of packages are cached and not upgraded. Docker cannot cache anything in layers after a dynamic variable has been used

Also bumping version of node and fixing test compilation issue


psoleckimoj (approved):
LGTM!

PR #75, opened 2022-03-22
Setup prometheus metrics by default.

dazoakley:
This change sets up prometheus metrics to be available on port 3001, and
with the helm chart changes they will automatically get scraped and be
available for alerts and dashboards in grafana.

The added metrics include:

- General nodejs stats: memory use, gc etc
- HTTP server requests: counters and timings of all served HTTP requests
  by the app.
- HTTP client requests: counters and timings of all HTTP requests to
  other upstream APIs (as long as they are based off `restClient.ts`).
- Upstream healthchecks: guages recording the status/health of each
  upstream service when the healthcheck is tested.

This is all backported from the `manage-recalls-ui` app, please let me
know what you think. :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,7 +63,7 @@ COPY --from=build --chown=appuser:appgroup \
 COPY --from=build --chown=appuser:appgroup \
         /app/node_modules ./node_modules
 
-EXPOSE 3000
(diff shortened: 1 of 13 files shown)


andrewrlee (approved):
LGTM

PR #81, opened 2022-04-20
Stop metrics test from hitting a real service and occasionally timing out

jonnywyatt:


andrewrlee (approved):
LGTM... I was wondering about this

PR #86, opened 2022-05-11
SLM-245 Restore cache prior to running the app for the integration tests

mikehalmamoj:
This caused an issue with our build where we received a segmentation fault as soon as the integration tests called the node app. Segmentation faults generally indicate an issue with one of the native C/C++ modules and it appears that one of these modules was relying on something we have stashed in the cache (is my best guess!)


andrewrlee (approved):
LGTM

PR #89, opened 2022-05-13
SDI-88: ♻️  Tidy up mocks and switch to multiplatform builds

petergphillips:


docker-compose-test.yml
```diff
@@ -2,16 +2,14 @@ version: '3.1'
 services:
 
   redis:
-    image: 'bitnami/redis:5.0'
+    image: 'redis:6.2'
```
petergphillips, line 5: switch to images that are suitable for `arm64`

integration_tests/mockApis/auth.ts
```diff
@@ -153,8 +153,8 @@ const stubUserRoles = () =>
 
 export default {
   getSignInUrl,
-  stubPing: (): Promise<[Response, Response]> => Promise.all([ping(), tokenVerification.stubPing()]),
```
petergphillips, line 156: this was a bit odd since the health check was calling both auth and token verification anyway

andrewrlee (approved):
lgtm

PR #95, opened 2022-05-20
Minor "code smell" fixes suggested by Sonar Cloud

ushkarev:


server/data/hmppsAuthClient.ts
```diff
@@ -55,7 +55,7 @@ export default class HmppsAuthClient {
   getUserRoles(token: string): Promise<string[]> {
     return HmppsAuthClient.restClient(token)
       .get({ path: '/api/user/me/roles' })
-      .then(roles => (<UserRole[]>roles).map(role => role.roleCode)) as Promise<string[]>
+      .then(roles => (<UserRole[]>roles).map(role => role.roleCode))
```
ushkarev, line 58: `roleCode` is already known to be a string because of the `<UserRole[]>` type assertion earlier

server/services/healthCheck.test.ts
```diff
@@ -36,22 +36,18 @@ describe('Healthcheck', () => {
...
 function erroredCheck(name: string): HealthCheckService {
   return () =>
-    new Promise((resolve, _reject) => {
-      resolve({
-        name: `${name}`,
-        status: 'ERROR',
-        message: 'some error',
-      })
+    Promise.resolve({
+      name: `${name}`,
```
petergphillips, line 49: surprised it couldn't work out that either `name: name` or simply `name` would work here

petergphillips (approved):
LGTM

server/services/userService.ts
```diff
@@ -11,6 +11,6 @@ export default class UserService {
 
   async getUser(token: string): Promise<UserDetails> {
     const user = await this.hmppsAuthClient.getUser(token)
-    return { ...user, displayName: convertToTitleCase(user.name as string) }
```
ushkarev, line 14: user is known to be a `User` so it's `name` property is already a string

server/services/healthCheck.test.ts
ushkarev: oh good spot! you should have waited to merge and i'd have changed that too :)

PR #110, opened 2022-07-22
Remove duplicate <main> elements

richpjames:
The `index.njk` and `error.njk` files use `layout.njk` which extends `govuk/template.njk`. `govuk/template.njk` already includes a `<main>` element.

[According to the HTML spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element) there should only be one `<main>` element present in the document at a time. Having multiple `<main>` elements could affect screen readers and other assistive technologies.

(NB the link in the [contributing guidelines](https://github.com/ministryofjustice/.github/blob/23444d612e846ceca655dc18c2a040c56a9f2f80/CONTRIBUTING.md) to [GDS guidance on commit hygene](https://gds-way.cloudapps.digital/standards/source-code.html#commit-messages) is dead so please let me know if there any improvements that need to be made to the commit message)

### DIFF
--- a/server/views/pages/error.njk
+++ b/server/views/pages/error.njk
@@ -1,13 +1,12 @@
 {% extends "../partials/layout.njk" %}
 
 {% set pageTitle = applicationName + " - Error" %}
+{% set mainClasses = "app-container govuk-body" %}
 
 {% block content %}
 
-  <main class="app-container">
-    <h1>{{ message }}</h1>
-    <h2>{{ status }}</h2>
-    <pre>{{ stack }}</pre>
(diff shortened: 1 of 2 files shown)


andrewrlee (approved):
LGTM

andrewrlee:
Thanks @richpjames!

Good spot - we had this infest one of our code bases and didn't realise this was the source

richpjames:
No probs @andrewrlee we had the same so thought it was best to fix the problem at source :) 
Thanks for reviewing and merging

PR #111, opened 2022-08-11
Bumping dependencies and fixing page width

andrewrlee:
There seems to be a lot of variability in page width so going with something that seems most popular in HMPPS


psoleckimoj (approved):
LGTM!

psoleckimoj (approved):
LGTM!

PR #112, opened 2022-08-12
 Add a `cspNonce` to the webSecurity setup

pezholio:
Based on what I’ve seen elsewhere, this seems to now be a common approach to allow us to inline scripts, see:

https://content-security-policy.com/nonce

The GOV.UK frontend has now been updated to support the use of the `cspNonce` local - see:

 https://github.com/alphagov/govuk-frontend/commit/2e40d744af6e6e4213ebc47644982d4eb94422d4

So we no longer need to add the inline hash, which is vulnerable to if the code in the frontend template is changed.

I’ve also removed the domain-specific overrides for jQuery scripts and styles, as we can use the nonce for this too.


brightonsbox (approved):
LGTM

PR #116, opened 2022-09-12
Speeding up jest tests

andrewrlee:
This speeds up the running of jest tests by enabling isolatedModules which has the effect of [disabling typechecking](https://kulshekhar.github.io/ts-jest/docs/getting-started/options/isolatedModules)

It also drastically reduces memory usage, allowing for running tests in parallel locally at least.

On my laptop this reduces the time to run the tests in this project from ~14 seconds to ~4 seconds. On larger projects the effect is much more pronounced, welcome-people-to-prison reduces build time from ~2mins, 20 seconds to ~25 seconds. In circle we still need to run in band but this is still significantly faster than before, in WPIP it reduces the build by over 1 min.

Type checking is still available in the IDE, it is also part of the husky pre-commit hook and run by circle as part of the build, so the risk of type errors slipping through are very small.

(We could possibly add a typechecking stage before running jest and it would still be much faster but not adding unless it becomes apparent that we need it )

It would be worth to re-assess this after jest 29 as there seems to be some fixes around a [memory issue](https://github.com/facebook/jest/issues/11956) that is part of node in versions > 16.10

NB: This will not work if type declaration (`d.ts`) files contain enums or any other constructs that generate javascript code. This seems to be a bit of an anti-pattern anyway (see [here](https://lukasbehal.com/2017-05-22-enums-in-declaration-files/)). Other tooling such as cypress will only allow you to import types from these files.


package.json
```diff
@@ -21,7 +21,8 @@
     "record-build-info": "node ./bin/record-build-info",
     "lint": "eslint . --cache --max-warnings 0",
     "typecheck": "tsc && tsc -p integration_tests",
-    "test": "jest --runInBand",
+    "test": "jest",
```
andrewrlee, line 24: run in parallel for local builds, but in band on circle

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-NODE_ENV=dev && node_modules/.bin/lint-staged && node_modules/.bin/tsc && npm test
+NODE_ENV=dev && node_modules/.bin/lint-staged && npm run typecheck && npm test
```
andrewrlee, line 4: include cypress tests typechecking pre commit hook

timharrison-moj (approved):
LGTM

PR #120, opened 2022-09-23
Fixing logging

andrewrlee:
There was an issue where we weren't sending trace info to app insights. This is because appInsights needs to be imported before bunyan is imported so it can do its instrumentation magic.

There was a related issue that obscured this. It was previously impossible to test app insights locally as dotenv wasn't set up correctly - it needed to happen before app insights is imported or app insights would prevent the app starting up.

So this moves dotenv to dev dependencies and preloads it before running the app via start:dev.

This removes some code that is only relevant for local development. It also means the application runs similar locally to how it would run in docker or kubernetes - it just expects the environment variables to be present.

Also moving the app insights import so it's very apparent that it's the first thing that happens when the app starts.


server/utils/azureAppInsights.ts
```diff
@@ -15,9 +14,7 @@ function version(): string {
 }
 
 export function initialiseAppInsights(): void {
-  // Loads .env file contents into | process.env
-  config()
-  if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY) {
+  if (process.env.APPLICATIONINSIGHTS_CONNECTION_STRING) {
```
andrewrlee, line 17: corrected this to right variable name

psoleckimoj (approved):
LGTM

PR #123, opened 2022-10-05
Adding badges

andrewrlee:
Adding the circle-ci and moj-compliant badges and adjusting the README so new projects will inherit these by default


package.json
```diff
@@ -33,7 +33,6 @@
     "npm": "^8"
   },
   "jest": {
-    "preset": "ts-jest",
```
andrewrlee, line 36: don't think this is needed any more?

psoleckimoj: It's a base config? We are pointing at the `ts-jest` module? https://jestjs.io/docs/configuration#preset-string

psoleckimoj (dismissed):
LGTM!

README.md
```diff
@@ -1,4 +1,7 @@
 # hmpps-template-typescript
+[![repo standards badge](https://img.shields.io/badge/dynamic/json?color=blue&style=for-the-badge&logo=github&label=MoJ%20Compliant&query=%24.result&url=https%3A%2F%2Foperations-engineering-reports.cloud-platform.service.justice.gov.uk%2Fapi%2Fv1%2Fcompliant_public_repositories%2Fhmpps-template-typescript)](https://operations-engineering-reports.cloud-platform.service.justice.gov.uk/public-github-repositories.html#hmpps-template-typescript "Link to report")
```
neilmendum, line 2: Not a big thing but in https://github.com/ministryofjustice/hmpps-incentives-api/pull/63 I used `style=flat` so the badge was the same size & shape as others

andrewrlee: ah that looks better! @neilmendum are you ok to merge this if you're happy with it?

neilmendum (approved):
LGTM

package.json
andrewrlee: I spent ages reading the docs and I still don't know. The tests run successfully without it

PR #134, opened 2022-12-19
Add rate limiting to auth endpoints

pezholio:
Following a recommendation by CodeQL, I’ve added some rate limiting to the endpoints that use authentication. This uses `express-rate-limit` and `rate-limit-redis` to prevent too many requests to specific endpoints/middleware. The rateLimiter function has options we can pass to set the window and max number of requests that can be made to an endpoint before hitting a rate limit.

For the authentication endpoints, I’ve set this to 10 requests per IP address per 10 minute window, and 100 requests per 1 minute windw for the authorisation endpoints which seems like a reasonable default to start with.


server/app.ts
```diff
@@ -34,10 +42,21 @@ export default function createApp(services: Services): express.Application {
...
+    setUpAuthentication(),
+    rateLimiter({ windowMs: authenticationRateLimitWindow, maxRequests: authenticationRateLimitMaxRequests }),
+  )
+  app.use(
+    authorisationMiddleware(),
+    rateLimiter({ windowMs: authorisationRateLimitWindow, maxRequests: authorisationRateLimitMaxRequests }),
+  )
+  app.use(
+    setUpCurrentUser(services),
+    rateLimiter({ windowMs: authorisationRateLimitWindow, maxRequests: authorisationRateLimitMaxRequests }),t
```
andrewrlee, line 56: typo here: `t`

pezholio: Whoops! I had some Prettier weirdness going on in my IDE, and rather than diagnose it, I made the changes in TextEdit 🤦

server/app.ts
```diff
@@ -34,10 +42,21 @@ export default function createApp(services: Services): express.Application {
   app.use(setUpWebRequestParsing())
   app.use(setUpStaticResources())
   nunjucksSetup(app, path)
-  app.use(setUpAuthentication())
-  app.use(authorisationMiddleware())
+
+  app.use(
+    setUpAuthentication(),
+    rateLimiter({ windowMs: authenticationRateLimitWindow, maxRequests: authenticationRateLimitMaxRequests }),
```
andrewrlee, line 48: Shouldn't this be before the authentication middleware to have effect?

server/app.ts
```diff
@@ -34,10 +42,21 @@ export default function createApp(services: Services): express.Application {
...
+  )
+  app.use(
+    authorisationMiddleware(),
+    rateLimiter({ windowMs: authorisationRateLimitWindow, maxRequests: authorisationRateLimitMaxRequests }),
```
andrewrlee, line 52: as you're setting it at the application layer you would only need to add it once but it would affect all route/middleware after you've hooked it in. 

The examples in the doc are adding different rate limiters to different paths - do you just want to do it on the auth ones?

pezholio: My concern (and the issue that was raised by CodeQL) was that the calls to `isAuthenticated` in the authorisationMiddleware are potentially expensive, so need some degree of rate limiting, hence why the authorisation rate limits are more permissive.

server/app.ts
```diff
@@ -16,10 +16,18 @@ import setUpStaticResources from './middleware/setUpStaticResources'
...
 
 import routes from './routes'
 import type { Services } from './services'
 
+// Add rate limits for each environment
+const authenticationRateLimitWindow = process.env.NODE_ENV === 'production' ? 10 * 60 * 1000 : 60
+const authenticationRateLimitMaxRequests = process.env.NODE_ENV === 'production' ? 10 : 100
+
+const authorisationRateLimitWindow = process.env.NODE_ENV === 'production' ? 60 * 1000 : 60
+const authorisationRateLimitMaxRequests = process.env.NODE_ENV === 'production' ? 60 * 1000 : 60
```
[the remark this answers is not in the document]
pezholio: Good call 👍

server/app.ts
andrewrlee: I don't think this is how [app.use](https://expressjs.com/en/4x/api.html#app.use) works. 

you have said

```
app.use(middleware1, middleware2)
app.use(middleware3, middleware4)
```

which is going to mean every requests will hit the middleware in the order:

middleware1 -> middleware2 -> middleware3 -> middleware4

so you will have no protection on middleware1 (`setUpAuthentication`) and then every request will go through 2 separate middleware requests.

andrewrlee: the examples in the express-rate-limiter docs show adding rate limiter to specific paths

helm_deploy/hmpps-template-typescript/values.yaml
```diff
@@ -32,10 +32,14 @@ generic-service:
...
-    TOKEN_VERIFICATION_ENABLED: "true"
-    APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/"
+    NODE_ENV: 'production'
+    REDIS_TLS_ENABLED: 'true'
+    TOKEN_VERIFICATION_ENABLED: 'true'
+    APPLICATIONINSIGHTS_CONNECTION_STRING: 'InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/'
+    AUTHENTICATION_WINDOW_TIME_IN_MINUTES: '10'
+    AUTHENTICATION_MAX_REQUESTS: '10'
+    AUTHORISATION_WINDOW_TIME_IN_MINUTES: '1'
+    AUTHORISATION_MAX_REQUESTS: '100'
```
andrewrlee, line 42: maybe a `RATE_LIMITER` prefix would help give context - or maybe you could remove from the helm chart and just let it default in config.ts?

server/app.ts
pezholio: Ah ha, I think I've got it now 👍

server/middleware/authorisationMiddleware.ts
```diff
@@ -14,7 +17,12 @@ export default function authorisationMiddleware(authorisedRoles: string[] = []):
         return res.redirect('/authError')
       }
 
-      return next()
+      return next(
+        rateLimiter({
+          windowMs: config.rateLimits.authorisation.windowTime,
+          maxRequests: config.rateLimits.authorisation.maxRequests,
+        }),
+      )
```
andrewrlee, line 25: This will just pass the middleware as a function to the error handler

I think this only compiles as next can take `any` as an error object

server/middleware/setUpAuthentication.ts
```diff
@@ -49,5 +50,12 @@ export default function setUpAuth(): Router {
     next()
   })
 
+  router.use(
+    rateLimiter({
+      windowMs: config.rateLimits.authentication.windowTime,
+      maxRequests: config.rateLimits.authentication.maxRequests,
+    }),
+  )
```
andrewrlee, line 58: this will still have the same problem as before, the router is not attached to any path and there is no isolation between routers - so by adding the router here you'll be adding it to every request.

look at the following for example: 
```
const express = require('express')
const app = express()

const router = express.Router()
app.use(router)

router.use((req, res, next) => {
  res.locals.msg = 'router 1'
  next()
})

app.get('/', (req, res) => {  res.send('Hello from: ' + res.locals.msg)})

app.listen(3000, () => { console.log(`Example app listening`) })
```
Hitting the endpoint will return `Hello from: router 1`

Also order really matters, adding middleware later will mean its visited later.
So in this case the rate limiter only kicks in after passport handles any requests.

pezholio: So, given that the only paths we have here are the sign in / out etc paths, is it fair to say they're the only ones we can/should ratelimit? Otherwise we're ratelimiting the entire stack?

andrewrlee: yes, I think so

andrewrlee:
Closing this as we wouldn't want to rate limit by ip address. Needs more thought as we have lots of users and few IP addresses

PR #135, opened 2022-12-20
Fix conflicting prettier / eslint rule

pezholio:
In eslint, we ask for a trailing comma, while the prettier rules are set to `es5`. This causes issues if you have your IDE set up to fix on save, as one linter kicks in before the other, causing conflicting fixes.


andrewrlee (approved):
LGTM

PR #137, opened 2022-12-29
Force setting of slack channels when customising project

andrewrlee:


tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compileOnSave": true,
   "compilerOptions": {
-    "target": "es2018",
+    "lib": ["ES2021"],
+    "target": "es2021",
```
andrewrlee, line 5: Bumping this to the latest compatible version with node 18

ushkarev: alternatively, it might be easier to use [@tsconfig/node18](https://www.npmjs.com/package/@tsconfig/node18) which actually sets `lib` and `target` to "es2022"…?

mattops (approved):

PR #149, opened 2023-02-03
Revert build_multiplatform_docker

neilmendum:
because it causes the build to take over an hour

[See Slack discussion](https://mojdt.slack.com/archives/C69NWE339/p1671529301455009?thread_ts=1671529075.740459&cid=C69NWE339)


andrewrlee (approved):
LGTM

PR #158, opened 2023-03-01
Point to correct locations for applicationVersion

richpjames:
The generated build-info.json and package.json files are in the parent directory.  
This was causing issues for us in Approved Premises so I thought it would be worthwhile fixing upstream.


richpjames:
I've noticed our repo has a different structure to those that come 'out of the box' from this template so I don't think this contribution would be helpful for the majority of use cases.

PR #170, opened 2023-03-20
Add HMPPS Auth URL to form-action CSP string

bryangaledxw:
Update the Content Security Policy to allow the HMPPS Auth URL as a possible form action target.

Currently, if a 403 error occurs on a GET request, this will be captured by the error handling setup in errorHandler.ts, and the user will be redirected to the sign out URL, which then redirects to the HMPPS Auth URL. However, if a 403 error occurs on a POST request, this second redirect may not occur, and the user may, depending on their choice of browser, be frozen on the form page they just submitted.

Due to CSP implementation details that vary between browsers, adding the HMPPS Auth URL to our form action targets allows this second redirect to occur as expected.

Note that several projects have already implemented a similar change:
- https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/7c9997ade13f599e3c8cac0d92ac176a097edff8/server/middleware/setUpWebSecurity.ts#L20
- https://github.com/ministryofjustice/make-recall-decision-ui/blob/e09b64f52eb908f2a9f43ca8e0df13c4202e7c23/server/middleware/setUpWebSecurity.ts#L31
- https://github.com/ministryofjustice/hmpps-incentives-ui/blob/b601aea8900735a78b003c623083d35f4b8b19e9/server/middleware/setUpWebSecurity.ts#L22

For further context, see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/form-action


andrewrlee (approved):
LGTM

PR #172, opened 2023-03-28
Removing unnecessary build

andrewrlee:
All 3 processes: tsc, sass and copy-views are run by concurrently at start up anyway


psoleckimoj (approved):
LGTM!

PR #178, opened 2023-04-13
SDIT-738: ⚡️ Cache static resources for 1 hour

petergphillips:


server/config.ts
```diff
@@ -32,7 +32,7 @@ export interface ApiConfig {
 export default {
   production,
   https: production,
-  staticResourceCacheDuration: 20,
```
petergphillips, line 35: this set the cache headers to cache for 20 seconds, which seems pointless

andrewrlee (approved):
LGTM

PR #188, opened 2023-05-09
Tie css cache to version of application

andrewrlee:
At the moment the cache is linked to the start up time of pod, so get unnecessary cache misses for every client, for each pod in the cluster and also as pods restart

This ties the cache to the git short hash of the deployment rather than having a cached version for each pod


louisenorris (approved):
LGTM

PR #190, opened 2023-05-12
Fix version not appearing in application insights

andrewrlee:
This previously relied on running a shell script to generate a file with a json payload in it.

The code that read this file to extract out the version for the cache improvement and also setting the application version in app insights, was looking in the wrong location

There was another location that looked up the file and read in the details for the health endpoint which was looking in the right place

This change moves to reading the version and git reference into an env var in the docker file instead, which means we can centralise how this info is made available and remove the additional file management

This should be a safe fix as the build info file was previously being generated from the docker build anyway - so the git sha ref should be available


server/applicationInfo.ts
```diff
@@ -0,0 +1,12 @@
+import fs from 'fs'
+import path from 'path'
+import config from './config'
+
+const packageJson =
+  process.env.NODE_ENV !== 'test' ? path.join(__dirname, '../../package.json') : path.join(__dirname, '../package.json')
```
andrewrlee, line 6: not very happy with the check to see if we're running a test or not.

Problem is that ts-jest transpiles js on the fly and the running dir is the ts source folder, whereas when running the application we're one layer down (in the transpilied code in the dist folder)

andrewrlee: tidied this up in later commit

server/utils/nunjucksSetup.ts
```diff
@@ -1,13 +1,13 @@
...
 import express from 'express'
-import * as pathModule from 'path'
+import path from 'path'
 import { initialiseName } from './utils'
-import applicationVersion from '../applicationVersion'
+import { ApplicationInfo } from '../applicationInfo'
 
 const production = process.env.NODE_ENV === 'production'
 
-export default function nunjucksSetup(app: express.Express, path: pathModule.PlatformPath): void {
```
andrewrlee, line 10: passing `path` in was unnecessary

psoleckimoj (approved):
LGTM!

PR #197, opened 2023-06-09
Do not retry POST requests by default

andrewrlee:
It doesn't really make sense to retry non-idempotent calls - there is a risk of double creating resources.

Also moving sanitised error over to a real error rather than an object. Makes it a little bit easier to test these: `expect(..).reject.throws` etc.. doesn't work if you don't have real errors


psoleckimoj (approved):
LGTM!

PR #198, opened 2023-06-09
Adding changelog

andrewrlee:
This is to assist teams following along with improvements that get back ported to the template project.

Provides a place where we can list major improvements along with some context as to what benefits these change make.

Went for a lo-fi approach, keep it simple and we can evolve or even drop the idea if we think we should


louisenorris (approved):
LGTM

PR #201, opened 2023-07-03
Mock out appInfo as used in multiple locations

psoleckimoj:
On a project cloned from the template I ran into issues whereby the proper `applicationInfo` was being called rather than a mocked value. As it's now used in multiple locations, just mocking out the value for `nunjucksSetup` isn't enough.

This mocks out the include for testing.

Thoughts?


tpmcgowan (approved):
LGTM

server/routes/testutils/appSetup.ts
```diff
@@ -9,12 +9,19 @@ import * as auth from '../../authentication/auth'
...
+    default: jest.fn(() => {
+      return {
+        applicationName: 'test',
+        buildNumber: '1',
+        gitRef: 'long ref',
+        gitShortHash: 'short ref',
+      } as ApplicationInfo
+    }),
+  }
+})
```
andrewrlee, line 24: mmmh... why can't you inject the application info into the place where you're using it?
 
you should be able to pass it in along with the service overrides [here](https://github.com/ministryofjustice/hmpps-template-typescript/pull/201/files#diff-47e00f44a4610e3f7bcaad8433c7bccf2ae050a627a3399de9553d7c910cd7b0R68)

andrewrlee: Have you got an example of where it was causing you an issue?

psoleckimoj: I'm sure I tried that but didn't seem to work. I'll give it another go. Just saves having to do that in every test file, and the build info is never going to want pulling from `package.json` in tests?

psoleckimoj: @andrewrlee in this PR (it's actually Draft), when trying to mock out the routes I think.... https://github.com/ministryofjustice/hmpps-developer-portal/pull/2/files

PR #202, opened 2023-07-04
Temp fix for semver vuln

psoleckimoj:
Can be checked on each update by removing the override and seeing if it's still needed.


andrewrlee (approved):
LGTM

PR #223, opened 2023-08-30
Fix CircleCI workflows for cypress

ushkarev:


.circleci/config.yml
```diff
@@ -52,7 +52,7 @@ jobs:
             - node_modules
             - build
             - dist
-            - .cache/Cypress
+            - assets/stylesheets
```
ushkarev, line 55: If compiled stylesheets aren't available for integration tests, some elements end up having entirely incorrect layouts and fail. Additionally, there is no `.cache` folder in `.` – the Cypress cache this tries to persist is in `~` (and that is already cached by a previous `save_cache` step

Mjwillis (approved):

PR #231, opened 2023-09-21
HEAT-106: Standardise endpoints

psoleckimoj:
Attempt to bring health and info endpoints inline with the Kotlin template.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7.2.1
+  hmpps: ministryofjustice/hmpps@7.3.4
   slack: circleci/slack@4.12.5
 
 parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7.3.4
+  hmpps: ministryofjustice/hmpps@7
   slack: circleci/slack@4.12.5
 
 parameters:
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,12 @@ FROM node:18.17-bullseye-slim as base
 
 ARG BUILD_NUMBER=1_0_0
 ARG GIT_REF=not-available
+ARG GIT_BRANCH=main
 
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
 ENV TZ=Europe/London
+ENV GIT_BRANCH ${GIT_BRANCH}
 RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezone
 
 RUN addgroup --gid 2000 --system appgroup && \
@@ -28,6 +30,8 @@ FROM base as build
 
 ARG BUILD_NUMBER=1_0_0
 ARG GIT_REF=not-available
+ARG GIT_BRANCH=main
+ENV GIT_BRANCH ${GIT_BRANCH}
 
 RUN apt-get update && \
         apt-get install -y make python g++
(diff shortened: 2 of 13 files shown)


tpmcgowan:
Managed to get this the integration tests to pass with the following changes to two files (it wouldn't let me  comment directly on the PR changes as they weren't bits you've touched.... 🙄 )

Looks like you've changed the json format for /health; so just updating for that. Although it now passes, probably worth checking that what's returned at http://localhost:3007/health is what you're intending!

```diff
diff --git a/integration_tests/e2e/health.cy.ts b/integration_tests/e2e/health.cy.ts
index afa496b..e0d54e5 100644
--- a/integration_tests/e2e/health.cy.ts
+++ b/integration_tests/e2e/health.cy.ts
@@ -26,8 +26,8 @@ context('Healthcheck', () => {
       cy.task('stubTokenVerificationPing', 500)
 
       cy.request({ url: '/health', method: 'GET', failOnStatusCode: false }).then(response => {
-        expect(response.body.checks.hmppsAuth).to.equal('OK')
-        expect(response.body.checks.tokenVerification).to.contain({ status: 500, retries: 2 })
+        expect(response.body.components.hmppsAuth.status).to.equal('UP')
+        expect(response.body.components.tokenVerification.status).to.contain({ status: 500, retries: 2 })
       })
     })
   })
diff --git a/integration_tests/mockApis/tokenVerification.ts b/integration_tests/mockApis/tokenVerification.ts
index 1b9d47a..17be392 100644
--- a/integration_tests/mockApis/tokenVerification.ts
+++ b/integration_tests/mockApis/tokenVerification.ts
@@ -11,7 +11,7 @@ export default {
       response: {
         status,
         headers: { 'Content-Type': 'application/json;charset=UTF-8' },
-        jsonBody: { status: 'UP' },
+        jsonBody: { status: 'DOWN' },
       },
     }),
   stubVerifyToken: (active = true): SuperAgentRequest =>
```

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7.2.1
+  hmpps: ministryofjustice/hmpps@7.3.4
```
andrewrlee, line 4: think dependabot may have introduced this very specific pinning - would probably better to just let it float at `7`.

Dockerfile
```diff
@@ -3,10 +3,12 @@ FROM node:18.18-bullseye-slim as base
 
 ARG BUILD_NUMBER=1_0_0
 ARG GIT_REF=not-available
+ARG GIT_BRANCH=main
 
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
 ENV TZ=Europe/London
+ENV GIT_BRANCH ${GIT_BRANCH}
```
andrewrlee, line 11: The ENV above has an `=` between key and value. I think this [needs it too](https://docs.docker.com/engine/reference/builder/#env)?

Dockerfile
```diff
@@ -28,6 +30,8 @@ FROM base as build
 
 ARG BUILD_NUMBER=1_0_0
 ARG GIT_REF=not-available
+ARG GIT_BRANCH=main
+ENV GIT_BRANCH ${GIT_BRANCH}
```
andrewrlee, line 34: same here

Dockerfile
```diff
@@ -3,10 +3,12 @@ FROM node:18.18-bullseye-slim as base
 
 ARG BUILD_NUMBER=1_0_0
 ARG GIT_REF=not-available
+ARG GIT_BRANCH=main
```
andrewrlee, line 6: might be better to default this to `not-available`? it'd be more obvious if its not been set

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
```
andrewrlee, line 1: Might be good to update CHANGELOG.md too?

andrewrlee (approved):
LGTM

PR #233, opened 2023-09-25
SDIT-1108: 🔧 Don't default build args

petergphillips:


Dockerfile
```diff
@@ -1,24 +1,29 @@
...
-ENV GIT_REF ${GIT_REF:-xxxxxxxxxxxxxxxxxxx}
+# Cache breaking and ensure required build / git args defined
+RUN test -n "$BUILD_NUMBER" || (echo "BUILD_NUMBER not set" && false)
+RUN test -n "$GIT_REF" || (echo "GIT_REF not set" && false)
+RUN test -n "$GIT_BRANCH" || (echo "GIT_BRANCH not set" && false)
+
+# Define env variables for runtime health / info
+ENV BUILD_NUMBER ${BUILD_NUMBER}
+ENV GIT_REF ${GIT_REF}
+ENV GIT_BRANCH ${GIT_BRANCH}
```
andrewrlee, line 26: do you need `=` or not? 

`ENV TZ=Europe/London`

but others don't

PR #235, opened 2023-09-27
SDIT-1108: ✨ Add in environment name to header

petergphillips:


andrewrlee (approved):
LGTM 

(we could do with adding something to the changelog for this)

PR #238, opened 2023-10-03
NON-270: Improve REST client

ushkarev:
REST client improvements:
- Adds commonly used PATCH, PUT and DELETE methods
- Allows passing query parameters to all methods (so you don't need to build a complex url path remembering to `encodeURIComponent`)
- Makes response body type generic (was always `unknown`)

Also, propagate user types into `res.locals` in request handlers.

Before:

```
 PASS  server/data/restClient.test.ts
  POST
    ✓ Should return response body
    ✓ Should return raw response body
    ✓ Should not retry by default
    ✓ retries if configured to do so
    ✓ can recover through retries
```

After:

```
 PASS  server/data/restClient.test.ts
  Method: get
    ✓ should return response body
    ✓ should return raw response body
    ✓ should retry by default
    ✓ can recover through retries
  Method: patch
    ✓ should return response body
    ✓ should return raw response body
    ✓ should not retry by default
    ✓ should retry if configured to do so
    ✓ can recover through retries
  Method: post
    ✓ should return response body
    ✓ should return raw response body
    ✓ should not retry by default
    ✓ should retry if configured to do so
    ✓ can recover through retries
  Method: put
    ✓ should return response body
    ✓ should return raw response body
    ✓ should not retry by default
    ✓ should retry if configured to do so
    ✓ can recover through retries
  Method: delete
    ✓ should return response body
    ✓ should return raw response body
    ✓ should retry by default
    ✓ can recover through retries
```

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -1,3 +1,5 @@
+import type { UserDetails } from '../../services/userService'
+
 export default {}
 
 declare module 'express-session' {
@@ -10,8 +12,7 @@ declare module 'express-session' {
 
 export declare global {
   namespace Express {
-    interface User {
-      username: string
+    interface User extends Partial<UserDetails> {
       token: string
       authSource: string
     }
@@ -20,6 +21,10 @@ export declare global {
       verified?: boolean
       id: string
       logout(done: (err: unknown) => void): void
+    }
+
+    interface Locals {
+      user: Express.User
     }
   }
 }
--- a/server/data/hmppsAuthClient.ts
+++ b/server/data/hmppsAuthClient.ts
@@ -1,4 +1,5 @@
 import { URLSearchParams } from 'url'
+
 import superagent from 'superagent'
 
 import type TokenStore from './tokenStore'
@@ -32,8 +33,14 @@ function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagen
 }
 
 export interface User {
-  name: string
-  activeCaseLoadId: string
+  username: string
+  name?: string
+  active?: boolean
+  authSource?: string
+  uuid?: string
+  userId?: string
+  staffId?: number // deprecated, use userId
+  activeCaseLoadId?: string // deprecated, use user roles api
 }
 
 export interface UserRole {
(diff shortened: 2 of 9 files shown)


server/data/hmppsAuthClient.ts
```diff
@@ -32,8 +33,14 @@ function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagen
...
-  name: string
-  activeCaseLoadId: string
+  username: string
+  name?: string
+  active?: boolean
+  authSource?: string
+  uuid?: string
+  userId?: string
+  staffId?: number // deprecated, use userId
+  activeCaseLoadId?: string // deprecated, use user roles api
```
ushkarev, line 43: cf. [hmpps-auth#UserDetail](https://github.com/ministryofjustice/hmpps-auth/blob/417c65f73d5f264d226cddb72057f92e53d054d6/src/main/kotlin/uk/gov/justice/digital/hmpps/oauth2server/model/UserDetail.kt)

server/@types/express/index.d.ts
```diff
@@ -21,5 +22,9 @@ export declare global {
       id: string
       logout(done: (err: unknown) => void): void
     }
+
+    interface Locals {
+      user: Express.User
+    }
```
ushkarev, line 28: The `setUpCurrentUser` middleware always populates `res.locals.user` via the user service

Mjwillis (approved):
Nice work!

server/data/hmppsAuthClient.ts
andrewrlee: slightly odd that some of these fields are prison specific - though activeCaseLoadId was here before.

andrewrlee: I guess auth ideally would expose this as a union type

server/data/restClient.ts
```diff
@@ -1,27 +1,24 @@
...
+import type { ApiConfig } from '../config'
 import type { UnsanitisedError } from '../sanitisedError'
 import { restClientMetricsMiddleware } from './restClientMetricsMiddleware'
 
-interface GetRequest {
-  path?: string
-  query?: string
+interface Request {
+  path: string
+  query?: object | string
```
andrewrlee, line 14: think this will need to be `Record<string, unknown>` otherwise you won't be able to pass anything other than empty object?

ushkarev: `Record<string, unknown>` "satisfies" `object`. i just used the official type accepted by the superagent function:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8139a88dd28ea580afedc673c18ceb24df95effb/types/superagent/index.d.ts#L166

psoleckimoj (approved):
LGTM!

PR #247, opened 2023-10-24
HAAR-1891: Update deprecated endpoints with new endpoints

ravmoj:
Update deprecated endpoints with new (manage users api) endpoints

### DIFF
--- a/feature.env
+++ b/feature.env
@@ -1,5 +1,6 @@
 PORT=3007
 HMPPS_AUTH_URL=http://localhost:9091/auth
+MANAGE_USERS_API_URL=http://localhost:9091
 TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
 TOKEN_VERIFICATION_ENABLED=true
 NODE_ENV=development
--- a/feature.env
+++ b/feature.env
@@ -1,6 +1,6 @@
 PORT=3007
 HMPPS_AUTH_URL=http://localhost:9091/auth
-MANAGE_USERS_API_URL=http://localhost:9091
+MANAGE_USERS_API_URL=http://localhost:9091/manage-users-api
 TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
 TOKEN_VERIFICATION_ENABLED=true
 NODE_ENV=development
--- a/integration_tests/mockApis/auth.ts
+++ b/integration_tests/mockApis/auth.ts
@@ -122,7 +122,7 @@ const stubUser = (name: string) =>
   stubFor({
     request: {
       method: 'GET',
-      urlPattern: '/auth/api/user/me',
+      urlPattern: '/users/me',
     },
     response: {
       status: 200,
@@ -142,7 +142,7 @@ const stubUserRoles = () =>
   stubFor({
     request: {
       method: 'GET',
-      urlPattern: '/auth/api/user/me/roles',
+      urlPattern: '/users/me/roles',
     },
     response: {
       status: 200,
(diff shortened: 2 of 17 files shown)


feature.env
```diff
@@ -1,5 +1,6 @@
 PORT=3007
 HMPPS_AUTH_URL=http://localhost:9091/auth
+MANAGE_USERS_API_URL=http://localhost:9091
```
brightonsbox, line 3: Wiremock serves all stubs from port 9091 regardless of which API it's stubbing.

We have a bit of a convention therefore, to distinguish different apis with a 'base':

e.g. for auth it's `http://localhost:9091/auth`, for token verification it's `http://localhost:9091/verification`...

so maybe for manage-users-api it needs to be `http://localhost:9091/manage-users-api`

integration_tests/mockApis/auth.ts
```diff
@@ -122,7 +122,7 @@ const stubUser = (name: string) =>
   stubFor({
     request: {
       method: 'GET',
-      urlPattern: '/auth/api/user/me',
+      urlPattern: '/users/me',
```
brightonsbox, line 125: [This](https://github.com/ministryofjustice/hmpps-template-typescript/pull/247/files#r1371550755) means that when stubbing you need to prefix these with `manage-users-api`

brightonsbox: And in fact because we've got separate files per API, might make sense pulling these stubs out into there own `mockApis/manageUsersApi.ts` file

server/config.ts
```diff
@@ -61,6 +61,18 @@ export default {
...
       systemClientSecret: get('SYSTEM_CLIENT_SECRET', 'clientsecret', requiredInProduction),
     },
+    manageUsersApi: {
+      url: get('MANAGE_USERS_API_URL', 'http://localhost:9091', requiredInProduction),
+      timeout: {
+        response: Number(get('MANAGE_USERS_API_TIMEOUT_RESPONSE', 10000)),
+        deadline: Number(get('MANAGE_USERS_API_TIMEOUT_DEADLINE', 10000)),
+      },
+      agent: new AgentConfig(Number(get('MANAGE_USERS_API_TIMEOUT_RESPONSE', 10000))),
+      apiClientId: get('API_CLIENT_ID', 'clientid', requiredInProduction),
```
brightonsbox, line 71: Don't need apiClientId, apiClientSecret, systemClientId, systemClientSecret here

server/data/manageUsersApiClient.ts
```diff
@@ -0,0 +1,38 @@
...
+  staffId?: number // deprecated, use userId
+  activeCaseLoadId?: string // deprecated, use user roles api
+}
+
+export interface UserRole {
+  roleCode: string
+}
+
+export default class ManageUsersApiClient {
+  constructor(private readonly tokenStore: TokenStore) {}
```
brightonsbox, line 22: Don't need a token store, this is only an auth specific thing, never used in the methods below

server/data/index.ts
```diff
@@ -19,8 +20,9 @@ type RestClientBuilder<T> = (token: string) => T
 export const dataAccess = () => ({
   applicationInfo,
   hmppsAuthClient: new HmppsAuthClient(new TokenStore(createRedisClient())),
+  manageUsersApiClient: new ManageUsersApiClient(new TokenStore(createRedisClient())),
```
brightonsbox, line 23: No separate token store required

brightonsbox (approved):
LGTM

PR #255, opened 2023-11-06
HAAR-1891: Add newly-included Manage Users api to health checks

ushkarev:
I think this was missed off from #247 when moving a couple of calls from HMPPS Auth to HMPPS Manage Users

### DIFF
--- a/integration_tests/mockApis/manageUsersApi.ts
+++ b/integration_tests/mockApis/manageUsersApi.ts
@@ -37,7 +37,19 @@ const stubUserRoles = () =>
     },
   })
 
+const ping = () =>
+  stubFor({
+    request: {
+      method: 'GET',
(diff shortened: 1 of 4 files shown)


integration_tests/mockApis/manageUsersApi.ts
```diff
@@ -37,7 +37,19 @@ const stubUserRoles = () =>
...
+    },
+    response: {
+      status: 200,
+    },
+  })
+
 export default {
   stubManageUser: (name = 'john smith'): Promise<[Response, Response]> =>
     Promise.all([stubUser(name), stubUserRoles()]),
+  stubManageUsersPing: ping,
```
ushkarev, line 54: the stubbing code is copied almost verbatim from auth

andrewrlee (approved):
LGTM

PR #256, opened 2023-11-07
Migrating project to use groups in allowlist

andrewrlee:
This PR migrates the project to use groups of IPs in their allowlist.

By referring to groups to IP addresses, we can centralize the definition of groups of ip addresses.
If these lists require changing in the future, we can change the definition once and future deploys across all services will automatically include these new IPs.

1 allowlist(s) have been detected that can be migrated.

### New Groups

The effect of applying this PR is as follows:

- The following groups will be applied: `internal`
- The size of the allowlist defined in this file will change: `8 => 0 (8 removed)`

### Added IPs

The new Group membership will result in the following IPs being added to your allowlist by applying this PR:

  Merging this PR should not result in any additional IP addresses being added to the allowlist.

### Removed IPs

The following IPs have been identified as unnecessary and will be removed by applying this PR:

- health-kick (35.177.252.195/32)

### DIFF
--- a/helm_deploy/hmpps-template-typescript/Chart.yaml
+++ b/helm_deploy/hmpps-template-typescript/Chart.yaml
@@ -5,7 +5,7 @@ name: hmpps-template-typescript
 version: 0.2.0
 dependencies:
   - name: generic-service
-    version: 2.6.5
+    version: "2.7"
     repository: https://ministryofjustice.github.io/hmpps-helm-charts
   - name: generic-prometheus-alerts
     version: 1.3.3
--- a/helm_deploy/hmpps-template-typescript/values.yaml
+++ b/helm_deploy/hmpps-template-typescript/values.yaml
@@ -1,18 +1,17 @@
----
 generic-service:
   nameOverride: hmpps-template-typescript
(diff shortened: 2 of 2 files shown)


helm_deploy/hmpps-template-typescript/Chart.yaml
```diff
@@ -5,7 +5,7 @@ name: hmpps-template-typescript
 version: 0.2.0
 dependencies:
   - name: generic-service
-    version: 2.6.5
+    version: "2.7"
```
petergphillips, line 8: should really go to 2.8

petergphillips (approved):
LGTM

helm_deploy/hmpps-template-typescript/values.yaml
```diff
@@ -54,14 +53,8 @@ generic-service:
       REDIS_AUTH_TOKEN: "auth_token"
 
   allowlist:
-    office: "217.33.148.210/32"
-    health-kick: "35.177.252.195/32"
```
ushkarev, line 58: fyi `health-kick` is not part of the "internal" group

PR #258, opened 2023-11-07
add manage users API to docker compose

Mindez:
Running locally is currently broken following the recent PR to add the manage users API as you get an ECONNREFUSED error due to the manage users API not running.

This PR adds the manage users API into the docker compose so it runs alongside other required local services (hmpps-auth, redis) which allows the template to run locally successfully again.


psoleckimoj (approved):
LGTM!

PR #265, opened 2023-11-15
SDIT-1223: 🎨 Minor cypress improvements

petergphillips:


integration_tests/e2e/signIn.cy.ts
```diff
@@ -41,6 +41,7 @@ context('SignIn', () => {
 
   it('User can manage their details', () => {
     cy.signIn()
+    cy.task('stubAuthManageDetails')
```
petergphillips, line 44: this was the only place that manage details was needed, so made sense to just call it here instead

integration_tests/mockApis/auth.ts
```diff
@@ -4,12 +4,14 @@ import { Response } from 'superagent'
 import { stubFor, getMatchingRequests } from './wiremock'
 import tokenVerification from './tokenVerification'
 
-const createToken = () => {
+const createToken = (roles: string[] = []) => {
+  // authorities in the session are always prefixed by ROLE.
+  const authorities = roles.map(role => (role.startsWith('ROLE_') ? role : `ROLE_${role}`))
```
petergphillips, line 9: helper so that we ensure tokens are always setup correctly

integration_tests/mockApis/auth.ts
```diff
@@ -95,7 +97,7 @@ const manageDetails = () =>
     },
   })
 
-const token = () =>
+const token = (roles: string[] = []) =>
```
petergphillips, line 100: allow roles to be passed through when token initially created

integration_tests/mockApis/auth.ts
```diff
@@ -120,6 +122,7 @@ const token = () =>
 export default {
   getSignInUrl,
   stubAuthPing: ping,
-  stubSignIn: (): Promise<[Response, Response, Response, Response, Response, Response]> =>
-    Promise.all([favicon(), redirect(), signOut(), manageDetails(), token(), tokenVerification.stubVerifyToken()]),
+  stubAuthManageDetails: manageDetails,
```
petergphillips, line 125: expose manage details separately as there was no need to stub it for a sign in

integration_tests/mockApis/manageUsersApi.ts
```diff
@@ -49,7 +47,7 @@ const ping = () =>
   })
 
 export default {
-  stubManageUser: (name = 'john smith'): Promise<[Response, Response]> =>
-    Promise.all([stubUser(name), stubUserRoles()]),
```
petergphillips, line 53: the manage user implementation doesn't call user roles, so there is no point stubbing it.  Instead we expose the stub below

andrewrlee (approved):
LGTM

PR #268, opened 2023-11-17
SDIT-1223: 🎨 Rename breadcrumb to be lowercase

petergphillips:


server/views/partials/breadcrumb.njk
```diff
@@ -1,14 +1,14 @@
 {% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
```
petergphillips, line 1: note that the govuk breadcrumb component is all lowercase here - see https://design-system.service.gov.uk/components/breadcrumbs/

ushkarev (approved):

PR #273, opened 2023-11-27
Use in-memory token store when developing locally

marcus-bcl:
Removes the need for a local Redis container

### DIFF
--- a/server/data/index.ts
+++ b/server/data/index.ts
@@ -14,7 +14,7 @@ import HmppsAuthClient from './hmppsAuthClient'
 import ManageUsersApiClient from './manageUsersApiClient'
 import { createRedisClient } from './redisClient'
 import RedisTokenStore from './tokenStore/redisTokenStore'
-import TestingTokenStore from './tokenStore/testingTokenStore'
+import InMemoryTokenStore from './tokenStore/inMemoryTokenStore'
 import config from '../config'
 
 type RestClientBuilder<T> = (token: string) => T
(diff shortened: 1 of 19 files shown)


server/data/index.ts
```diff
@@ -13,13 +13,17 @@ buildAppInsightsClient(applicationInfo)
...
+import TestingTokenStore from './tokenStore/testingTokenStore'
+import config from '../config'
 
 type RestClientBuilder<T> = (token: string) => T
 
 export const dataAccess = () => ({
   applicationInfo,
-  hmppsAuthClient: new HmppsAuthClient(new TokenStore(createRedisClient())),
+  hmppsAuthClient: new HmppsAuthClient(
+    config.redis.enabled ? new RedisTokenStore(createRedisClient()) : new TestingTokenStore(),
```
andrewrlee, line 25: Worth calling this InMemoryToken store? 
We use an in memory token store for jobs as it means don't need to worry about redis being accessible (dodgy DNS issues)

server/middleware/setUpWebSession.ts
```diff
@@ -1,19 +1,23 @@
...
 import RedisStore from 'connect-redis'
 import express, { Router } from 'express'
 import { createRedisClient } from '../data/redisClient'
 import config from '../config'
 import logger from '../../logger'
 
 export default function setUpWebSession(): Router {
-  const client = createRedisClient()
-  client.connect().catch((err: Error) => logger.error(`Error connecting to Redis`, err))
+  let store: Store
```
andrewrlee, line 10: should this fall back to the inmemory store?  what happens if you pass undefined?

server/data/index.ts
marcus-bcl: Good call, I'll rename it

server/middleware/setUpWebSession.ts
marcus-bcl: Express defaults to a MemoryStore, so passing undefined means it uses its own in-memory store for session data.

PR #274, opened 2023-11-28
Remove getUserRoles as an api call and add as decoded from the token

thomasridd:
### What

- Remove getUserRoles as an api call and get roles by decoding the token
- Add roles to the getUser(token) service method
- Update tests

### Why
This significantly reduces the number of calls needed to external services. In this case the api method getUserRoles(token) merely sends the token off to the manage-users-api which decodes it and sends back the result.

### DIFF
--- a/server/data/manageUsersApiClient.ts
+++ b/server/data/manageUsersApiClient.ts
@@ -28,10 +28,4 @@ export default class ManageUsersApiClient {
     logger.info('Getting user details: calling HMPPS Manage Users Api')
     return ManageUsersApiClient.restClient(token).get<User>({ path: '/users/me' })
   }
-
-  getUserRoles(token: string): Promise<string[]> {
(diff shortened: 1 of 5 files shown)


server/data/manageUsersApiClient.ts
```diff
@@ -28,10 +28,4 @@ export default class ManageUsersApiClient {
     logger.info('Getting user details: calling HMPPS Manage Users Api')
     return ManageUsersApiClient.restClient(token).get<User>({ path: '/users/me' })
   }
-
-  getUserRoles(token: string): Promise<string[]> {
```
andrewrlee, line 32: was this not being used anywhere?

andrewrlee (approved):
LGTM

server/data/manageUsersApiClient.ts
andrewrlee: apparently not

PR #276, opened 2023-11-29
Add local user account for development

marcus-bcl:
Removes the need for running HMPPS Auth or Manage Users API containers locally


server/testutils/createUserToken.ts
```diff
@@ -1,14 +1,23 @@
 import jwt from 'jsonwebtoken'
 
-export default function createUserToken(authorities: string[]) {
-  const payload = {
+interface TokenPayload {
```
ushkarev, line 3: doesn't matter really, but a more complete version of this type is here for reference
https://github.com/ministryofjustice/hmpps-non-associations/blob/839749915bf26cd7df3504e7e6e74c7860fbce44/server/middleware/authorisationMiddleware.ts#L6-L13

marcus-bcl: Nice, thanks. I've updated the `TokenPayload` interface to match.

PR #277, opened 2023-12-04
Requre user input when excucting rename script to ensure slack alert channels are set correctly.

mattops:


andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #285, opened 2023-12-14
HAAR-2061: Remove deprecated filed

ravmoj:
1 ) Removed staffId (userId has same info),
2) activeCaseLoadId : can be derived from 'me/caseloads' endpoint in 'nomis-user-roles-api'

### DIFF
--- a/server/data/manageUsersApiClient.ts
+++ b/server/data/manageUsersApiClient.ts
@@ -9,8 +9,7 @@ export interface User {
   authSource?: string
   uuid?: string
   userId?: string
-  staffId?: number // deprecated, use userId
-  activeCaseLoadId?: string // deprecated, use user roles api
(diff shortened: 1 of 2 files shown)


SimonMitchellMOJ (approved):
LGTM

server/data/manageUsersApiClient.ts
```diff
@@ -9,8 +9,7 @@ export interface User {
   authSource?: string
   uuid?: string
   userId?: string
-  staffId?: number // deprecated, use userId
```
tpmcgowan, line 12: Does this mean the wiremock stub should also be updated to remove `staffId` too? Here: https://github.com/ministryofjustice/hmpps-template-typescript/blob/92d8a3d19239a08686bb5de137bccdfe6b929c18/integration_tests/mockApis/manageUsersApi.ts#L15

ravmoj: Thanks, removed from stub

SimonMitchellMOJ (approved):
LGTM

PR #291, opened 2023-12-19
Update the breadcrumb partial with display current page title toggle

BarryGeeMOJ:
Adds a toggle - defaulting to true. To include or leave off the current page title from the breadcrumb partial.


psoleckimoj (approved):
LGTM!

PR #303, opened 2024-01-26
Update MoJ frontend, dropping jQuery which is no longer required

ushkarev:
Might be worth noting that cypress still bundles jquery so it will remain available for integration tests

### DIFF
--- a/server/middleware/setUpStaticResources.ts
+++ b/server/middleware/setUpStaticResources.ts
@@ -21,17 +21,12 @@ export default function setUpStaticResources(): Router {
     '/node_modules/govuk-frontend/dist',
     '/node_modules/@ministryofjustice/frontend/moj/assets',
     '/node_modules/@ministryofjustice/frontend',
-    '/node_modules/jquery/dist',
   ).forEach(dir => {
     router.use('/assets', express.static(path.join(process.cwd(), dir), cacheControl))
   })
 
   Array.of('/node_modules/govuk_frontend_toolkit/images').forEach(dir => {
     router.use('/assets/images/icons', express.static(path.join(process.cwd(), dir), cacheControl))
-  })
-
-  Array.of('/node_modules/jquery/dist/jquery.min.js').forEach(dir => {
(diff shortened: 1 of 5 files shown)


server/middleware/setUpStaticResources.ts
```diff
@@ -21,19 +21,10 @@ export default function setUpStaticResources(): Router {
     '/node_modules/govuk-frontend/dist',
     '/node_modules/@ministryofjustice/frontend/moj/assets',
     '/node_modules/@ministryofjustice/frontend',
-    '/node_modules/jquery/dist',
   ).forEach(dir => {
     router.use('/assets', express.static(path.join(process.cwd(), dir), cacheControl))
   })
 
-  Array.of('/node_modules/govuk_frontend_toolkit/images').forEach(dir => {
-    router.use('/assets/images/icons', express.static(path.join(process.cwd(), dir), cacheControl))
```
petergphillips, line 30: did you mean to remove this too?

ushkarev: yup i explained in that commit that we don't even install the deprecated `govuk_frontend_toolkit` package

petergphillips (approved):
LGTM

yndajas:
Seems the sortable table component still needs jQuery by the way

ushkarev:
@yndajas hmm that's weird… they [announced](https://mojdt.slack.com/archives/CH5RUSB27/p1706260500746629) that the 2.1 update
> also removes a dependency on using jQuery within MOJ frontend code
… so i guessed they missed a component? i presume their intention is to actually remove the jquery dependency so it might be worth raising it with them?

yndajas:
> … so i guessed they missed a component? i presume their intention is to actually remove the jquery dependency so it might be worth raising it with them?

Odd! There's a still-open issue from 2021 about this: https://github.com/ministryofjustice/moj-frontend/issues/134. jQuery is definitely still being used on `main`: https://github.com/search?q=repo:ministryofjustice/moj-frontend+$.&type=code. Will ask about the status of the dependency on Slack

PR #305, opened 2024-01-30
Update Nunjucks component paths

yndajas:
`node_modules/govuk-frontend/dist/components/` was missing `govuk/`
before `components/`, however we don't really need both the higher-level
and lower-level paths here. For greater consistency, just using the
higher-level path to enable a single clearly-namespaced route to each
GOV.UK and MOJ Nunjucks component makes more sense. The lower-level path
is often unused on projects that use this template

Imports should therefore look like this:

```njk
{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
```

... and not this:

```njk
{% from "checkboxes/macro.njk" import govukCheckboxes %}
```

### DIFF
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -32,9 +32,7 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
     [
       path.join(__dirname, '../../server/views'),
       'node_modules/govuk-frontend/dist/',
-      'node_modules/govuk-frontend/dist/components/',
       'node_modules/@ministryofjustice/frontend/',
-      'node_modules/@ministryofjustice/frontend/moj/components/',
     ],
     {
       autoescape: true,


server/utils/nunjucksSetup.ts
```diff
@@ -32,7 +32,7 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
     [
       path.join(__dirname, '../../server/views'),
       'node_modules/govuk-frontend/dist/',
-      'node_modules/govuk-frontend/dist/components/',
+      'node_modules/govuk-frontend/dist/govuk/components/',
```
psoleckimoj, line 35: Should we also add it to line 34 above?

ushkarev: no because then import paths from the docs (e.g. `{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}`) would not work

yndajas: I'm not too sure how this works to be honest/why we need a higher-level directory and a lower-level one. This is roughly in line with the lines below though:

```ts
      'node_modules/@ministryofjustice/frontend/',
      'node_modules/@ministryofjustice/frontend/moj/components/',
```

ushkarev: in fact, what i do in our team's projects is only retain these 2 search paths so that there is only one consistent route to each template
• `'node_modules/govuk-frontend/dist/'`
• `'node_modules/@ministryofjustice/frontend/'`

psoleckimoj: @ushkarev @yndajas happy to go with ☝️

yndajas: I've pushed an update and credited you as co-author @ushkarev - hope that's okay!

psoleckimoj (approved):
LGTM!

server/utils/nunjucksSetup.ts
ushkarev: whoa whoa don't blame me :p

.node-version
```diff
@@ -0,0 +1 @@
+20.11.0
```
ushkarev, line 1: might i suggest `20` instead of such a specific version? cos there's nothing in the repo that needs this _exact_ version and it becomes yet another place to keep up-to-date with the latest numbers.

do note there's already https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/.nvmrc

yndajas: Didn't mean to commit this - already force-pushed it out!

ushkarev (approved):

.node-version
ushkarev: i <3 force-push

yndajas:
@psoleckimoj @ushkarev do either of you have merge permissions?

PR #227, opened 2023-09-05
add configurable implementation of shared header and footer

whitfield-mj:

PR #307, opened 2024-01-31
SDIT-1450: 🐛 Fix manage details link to allow user to return

petergphillips:


ushkarev (approved):
looks legit… but won't really be necessary once the "micro-frontend" components are integrated in the template. i think most downstream projects already use them for the fancy header/footer

ref https://github.com/ministryofjustice/hmpps-template-typescript/pull/227

PR #315, opened 2024-02-15
PI-1789 added ability to get parameterised paths in app insights for FE

anthony-britton-moj:
Adds ability to use parameterised names for requests in app insights rather than using the full url.


andrewrlee (approved):
LGTM

PR #327, opened 2024-03-12
HMPPS Audit setup

leejacobson-moj:
## Overview

Adds HMPPS Audit to the template.


.eslintrc.json
```diff
@@ -33,6 +33,8 @@
         "plugin:prettier/recommended"
       ],
       "rules": {
+        "no-shadow": "off",
+        "@typescript-eslint/no-shadow": ["error"],
```
leejacobson-moj, line 37: https://github.com/typescript-eslint/typescript-eslint/issues/2483

helm_deploy/values-dev.yaml
```diff
@@ -13,6 +13,7 @@ generic-service:
     MANAGE_USERS_API_URL: "https://manage-users-api-dev.hmpps.service.justice.gov.uk"
     TOKEN_VERIFICATION_API_URL: "https://token-verification-api-dev.prison.service.justice.gov.uk"
     ENVIRONMENT_NAME: DEV
+    AUDIT_ENABLED: "false"
```
leejacobson-moj, line 16: Disabled by default since it requires some configuration.

server/config.ts
```diff
@@ -84,6 +84,12 @@ export default {
       agent: new AgentConfig(Number(get('TOKEN_VERIFICATION_API_TIMEOUT_RESPONSE', 5000))),
       enabled: get('TOKEN_VERIFICATION_ENABLED', 'false') === 'true',
     },
+    audit: {
+      queueUrl: get('AUDIT_SQS_QUEUE_URL', 'http://localhost:4566/000000000000/mainQueue', requiredInProduction),
+      region: get('AUDIT_SQS_REGION', 'eu-west-2', requiredInProduction),
+      serviceName: get('AUDIT_SERVICE_NAME', 'UNASSIGNED', requiredInProduction),
+      enabled: get('AUDIT_ENABLED', 'false') === 'true',
+    },
```
leejacobson-moj, line 92: I'm not sure this should sit under apis. I noticed this was where the prisoner profile project was defining the config though so I put it here for now.

server/data/hmppsAuditClient.test.ts
```diff
@@ -0,0 +1,110 @@
+import { mockClient } from 'aws-sdk-client-mock'
```
leejacobson-moj, line 1: Using `aws-sdk-client-mock` for unit tests.

server/data/hmppsAuditClient.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        service: 'hmpps-service',
+        subjectId: 'subject123',
+        subjectType: 'exampleType',
+        correlationId: 'request123',
+        details: { extraDetails: 'example' },
+      }
+
+      expect(actualResponse).toEqual({ MessageId: '123' })
+
+      const actualMessageInput = sqsMock.call(0).args[0].input as SendMessageCommandInput
```
leejacobson-moj, line 52: Inspect the call to the SQS mock client.

server/data/hmppsAuditClient.ts
```diff
@@ -0,0 +1,70 @@
...
+      const messageResponse = await this.sqsClient.send(
+        new SendMessageCommand({ MessageBody: JSON.stringify(sqsMessage), QueueUrl: this.queueUrl }),
+      )
+
+      logger.info(`HMPPS Audit SQS message sent (${messageResponse.MessageId})`)
+
+      return messageResponse
+    } catch (error) {
+      logger.error('Error sending HMPPS Audit SQS message, ', error)
+      if (throwOnError) throw error
```
leejacobson-moj, line 66: A question was raised last week about what we should do in the unlikely event an audit event fails. Martin Ballhatchet suggested it would be best to show an error since it risks user actions going unaudited.

The reason for this parameter is so teams can decide whether audit errors should be thrown or simply logged.

An alternative approach here might be for the client to always throw an error and for teams to decide how they want to handle that exception in the service layer?

server/routes/index.ts
```diff
@@ -2,13 +2,16 @@ import { type RequestHandler, Router } from 'express'
...
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export default function routes(service: Services): Router {
+export default function routes({ auditService }: Services): Router {
   const router = Router()
   const get = (path: string | string[], handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
 
-  get('/', (req, res, next) => {
+  get('/', async (req, res, next) => {
+    await auditService.logPageView(Page.EXAMPLE_PAGE, { who: res.locals.user.username, correlationId: req.id })
```
leejacobson-moj, line 13: An example audit event.

server/services/auditService.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export default class AuditService {
+  constructor(private readonly hmppsAuditClient: HmppsAuditClient) {}
+
+  async logAuditEvent(event: AuditEvent) {
+    await this.hmppsAuditClient.sendMessage(event)
+  }
+
+  async logPageView(page: Page, eventDetails: PageViewEventDetails) {
```
leejacobson-moj, line 22: This method might not belong in the template. The thinking was that most services will probably want to audit page views and this provides them a way to do that out of the box.

helm_deploy/hmpps-template-typescript/values.yaml
```diff
@@ -1,5 +1,6 @@
 generic-service:
   nameOverride: hmpps-template-typescript
+  serviceAccountName: "UNASSIGNED" # the IRSA service account name for accessing AWS resources
```
andrewrlee, line 3: is this going to prevent the pod from starting in development when this is merged?

package.json
```diff
@@ -93,9 +93,11 @@
     ]
   },
   "dependencies": {
+    "@aws-sdk/client-sqs": "^3.529.1",
     "@ministryofjustice/frontend": "^2.1.1",
     "agentkeepalive": "^4.5.0",
     "applicationinsights": "^2.9.5",
+    "aws-sdk-client-mock": "^3.0.1",
```
andrewrlee, line 100: should this be a dev dependency?

server/config.ts
andrewrlee: I think it might be better to move it out - want to look at automatically registering health checks for apis automatically and this wouldn't fall into that pattern.

server/data/hmppsAuditClient.ts
andrewrlee: I think this is nice - defaulting to throw but allowing overriding.

server/routes/testutils/appSetup.ts
```diff
@@ -45,6 +49,10 @@ function appSetup(services: Services, production: boolean, userSupplier: () => E
     }
     next()
   })
+  app.use((req, res, next) => {
+    req.id = uuidv4()
```
andrewrlee, line 53: not a comment for this PR, but we should probably add this as a custom dimension in app insights

helm_deploy/hmpps-template-typescript/values.yaml
leejacobson-moj: I'm not 100% sure if it will fail to deploy, but it's a good point and I've commented it out for now.

If there's value in integrating with audit, it could be done, but it would mean providing a valid service account name which might be problematic when cloning the project if teams don't realise it needs to be updated.

It might be better if it's left as a comment with the "UNASSIGNED" value since I guess teams are looking for "UNASSIGNED" references to update already?

helm_deploy/hmpps-template-typescript/values.yaml
```diff
@@ -2,6 +2,9 @@ generic-service:
   nameOverride: hmpps-template-typescript
   productId: "UNASSIGNED" # productId for the product that this belongs too, i.e. DPS001, see README.md for details
 
+  # the IRSA service account name for accessing AWS resources
+  # serviceAccountName: "UNASSIGNED"
```
andrewrlee, line 6: if we called this hmpps-template-typescript then it would be renamed to the same name as the project which might be a sensible default

leejacobson-moj: Updated 👍

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #350, opened 2024-04-29
SDIT-1682: ⬆️ Upgrade node and helm charts

petergphillips:


.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ parameters:
 
   node-version:
     type: string
-    default: 20.11-browsers
+    default: 20.12-browsers
```
petergphillips, line 21: looks like the Dockerfile had been updated, but this one was missed in the upgrade 🤷

psoleckimoj: I gave up and in the Dockerfile I usually just do 20-browsers but that's not available in CircleCIs images that I can see.

psoleckimoj (approved):
LGTM!

PR #373, opened 2024-06-11
MAP-1255 Add accessibility testing to integration tests by default

danbenton-mojdt:


package.json
```diff
@@ -97,6 +97,7 @@
     "@ministryofjustice/frontend": "^2.1.3",
     "agentkeepalive": "^4.5.0",
     "applicationinsights": "^2.9.5",
+    "axe-core": "^4.7.2",
```
petergphillips, line 100: any reason for this to be a standard dependency instead of a dev dependency?

integration_tests/pages/page.ts
```diff
@@ -1,19 +1,53 @@
...
     return new constructor()
   }
 
-  constructor(private readonly title: string) {
-    this.checkOnPage()
+  constructor(
+    private readonly title: string,
+    private readonly skipA11y = false,
+  ) {
+    if (title) {
```
petergphillips, line 15: why the check here?  Isn't `title` always going to be set?

package.json
danbenton-mojdt: No, moved to dev dependencies

integration_tests/pages/page.ts
danbenton-mojdt: Copied from other project without thinking about it much - removed!

integration_tests/pages/page.ts
```diff
@@ -1,19 +1,51 @@
...
 
 export default abstract class Page {
   static verifyOnPage<T>(constructor: new () => T): T {
     return new constructor()
   }
 
-  constructor(private readonly title: string) {
+  constructor(
+    private readonly title: string,
+    private readonly skipA11y = false,
```
petergphillips, line 13: not a great fan of booleans in constructors - difficult to know what the `super('Sign in', true)` means without having a look at the implementation

integration_tests/pages/page.ts
```diff
@@ -1,19 +1,51 @@
...
+  ) {
     this.checkOnPage()
   }
 
   checkOnPage(): void {
     cy.get('h1').contains(this.title)
+    if (!(Cypress.env('SKIP_AXE') || this.skipA11y)) {
```
petergphillips, line 20: we've got a `SKIP_AXE` and a `skipA11y`, wondering if it is better to have the same name for both i.e. `skipAxe` or maybe even `SKIP_ACCESSIBILITY_CHECKS` and `skipAccessibilityChecks`

petergphillips (approved):

integration_tests/pages/page.ts
psoleckimoj: Would be easier as an object then at least you can have meaningful attribute names in there and then use a boolean.

PR #375, opened 2024-06-13
Added ESBuild for building app and bundling/minfiying assets

Thomas-Geraghty:
- Added ESBuild for building project
- Added support for minifying and bundling client side assets
- Replaced nodemon with chokidar for watching/building project assets

Note: The base for this PR is a branch where I've moved the assets folder to /server/assets, probably not necessary in retrospect. It makes this build process a bit tidier :)


server/assets/scss/application.scss
```diff
@@ -4,8 +4,7 @@ $path: "/assets/images/";
 $moj-page-width: 1170px;
 $govuk-page-width: $moj-page-width;
 
-@import 'govuk/all';
-@import 'moj/all';
-
+@import "govuk-frontend/dist/govuk/all";
+@import "@ministryofjustice/frontend/moj/all";
 @import './components/header-bar';
 @import './local';
```
psoleckimoj, line 10: Missing newline

psoleckimoj (approved):
LGTM!

esbuild/assets.config.js
```diff
@@ -0,0 +1,49 @@
...
+
+const buildAssets = buildConfig => {
+  return esbuild.build({
+    entryPoints: buildConfig.assets.entryPoints,
+    outdir: buildConfig.assets.outDir,
+    entryNames: '[ext]/app',
+    minify: buildConfig.isProduction,
+    sourcemap: !buildConfig.isProduction,
+    platform: 'browser',
+    target: 'es2015',
```
ushkarev, line 27: tsconfig uses 2018 for node and a cursory look at caniuse.com makes me wonder if we can use 2018 here too? what do people think?

.eslintignore
```diff
@@ -4,3 +4,4 @@ assets
 cypress.json
 reporter-config.json
 dist/
+esbuild
```
ushkarev, line 7: wouldn't it be better to lint the esbuild config too? eslint complains about some things like console.log() statements and imports, but i imagine it wouldn't be tough to fix or ignore in place

package.json
```diff
@@ -6,18 +6,12 @@
...
   "scripts": {
     "prepare": "husky",
-    "copy-views": "cp -R server/views dist/server/",
-    "compile-sass": "sass --quiet-deps --no-source-map --load-path=node_modules/govuk-frontend/dist --load-path=node_modules/@ministryofjustice/frontend --load-path=. server/assets/scss/application.scss:./dist/server/assets/stylesheets/application.css --style compressed",
     "watch-ts": "tsc -w",
-    "watch-views": "nodemon --watch server/views -e html,njk -x npm run copy-views",
-    "watch-node": "DEBUG=gov-starter-server* nodemon -r dotenv/config --watch dist/ dist/server.js | bunyan -o short",
-    "watch-sass": "npm run compile-sass -- --watch",
-    "build": "npm run compile-sass && tsc && npm run copy-views",
+    "build": "NODE_ENV=production && node esbuild/esbuild.config.js --build",
```
ushkarev, line 10: why force the environment to always be production? the Dockerfile sets `NODE_ENV` so the deployed version would already end up compiling in production mode

server/views/partials/layout.njk
```diff
@@ -14,7 +14,5 @@
 {% endblock %}
 
 {% block bodyEnd %}
-  <script type="module" src="/assets/govukFrontendInit.js"></script>
```
ushkarev, line 17: the design system docs [say that you "must" use `type="module"`](https://frontend.design-system.service.gov.uk/import-javascript/) though it's probably not _that_ hard a requirement… will our compiled app.js load ok as a module?

esbuild/assets.config.js
psoleckimoj: I'd agree with a later version, even GDS has basically dropped IE support for example.

Dockerfile
```diff
@@ -53,9 +54,6 @@ COPY --from=build --chown=appuser:appgroup \
         /app/package-lock.json \
         ./
 
-COPY --from=build --chown=appuser:appgroup \
```
andrewrlee, line 56: presumably we don't need this as they're copied as part of dist?

assets/scss/application.scss
```diff
@@ -4,8 +4,8 @@ $path: "/assets/images/";
...
 $govuk-page-width: $moj-page-width;
 
-@import 'govuk/all';
-@import 'moj/all';
+@import "govuk-frontend/dist/govuk/all";
+@import "@ministryofjustice/frontend/moj/all";
 
 @import './components/header-bar';
-@import 'assets/scss/local';
+@import './local';
```
andrewrlee, line 11: meh.. missing new line

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

ushkarev:
@Thomas-Geraghty when cherry-picking this to our projects…
1) it was actually pretty easy to merge 🎉 
2) i noticed a few changes that might be worthwhile:
    - `nodemon` and `sass` _direct_ dev dependencies can be removed from package.json
    - there's no longer a need to persist `assets/stylesheets` in the CircleCI workflow nor git-ignore that path
    - (this is probs more personal style) swapping `console.log(…)` for `process.stderr.write(…)` and then we don't need the "no-console" eslint exemption
    - asset-building script logs caught exceptions, but app-building exited without info
    - i don't think static asset routes are needed to these paths: `/dist/assets/stylesheets` & `/dist/assets/js` – i know you just rewrote the existing ones to add the `/dist` prefix, but it feels tidier to not have multiple routes to the same files
    - i tried (and failed) to turn the build scripts into typescript, but instead [added some half-arsed type annotations](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/116/commits/ca371d41050dee4f1349e657e8c9640629449065) so at least an IDE helps even if `tsc` doesn't enforce things

which, if any, of these bullet points do you like and want me to put in here?

ReedSoftware:
@Thomas-Geraghty from our side, the only thing we've done is add an extra entry to `app.copy` array because our project structure is slightly different, but that was an easy 10 second job. I will echo your point `2.3` where certain errors will bomb out the server, which is a little annoying, but if you've got fixes that's great, as long as all errors that were bombing it out are all now caught. I just have one line in your PR I'm curious about (`return Promise.all([buildAssets(buildConfig), buildAdditionalAssets(buildConfig)])`) - this doesn't appear to have a catch and log or a catch and process.exit on it? It's on line 44.

Agree with all other points. Would be nice to have the config as ts - webpack and others manage it, and there appear to be examples of others using `.ts` files for esbuild config files, but type annotations would be a great first step.

ushkarev:
i’ve made the first pass at the above in [hmpps-template-typescript#389](https://github.com/ministryofjustice/hmpps-template-typescript/pull/389) and included your error-catching tweaks @Thomas-Geraghty

PR #377, opened 2024-06-17
Add hash-based cache busting for client side bundles

Thomas-Geraghty:
- Added `esbuild-plugin-manifest` module for generating a manifest file that maps original asset name to its generated output, structured like below
```
{
  "/assets/js/app.js": "/assets/js/app.UG7VY7MS.js",
  "/assets/css/app.css": "/assets/css/app.47MYFHW7.css"
}
```
- Added logic to read manifest file and use mappings through a nunjucks filter called `assetMap` to swap asset strings for the build output
- Removed old cache busting code

### DIFF
--- a/esbuild/assets.config.js
+++ b/esbuild/assets.config.js
@@ -1,13 +1,18 @@
+const path = require('node:path')
+
 const { copy } = require('esbuild-plugin-copy')
 const { sassPlugin } = require('esbuild-sass-plugin')
 const { clean } = require('esbuild-plugin-clean')
 const manifestPlugin = require('esbuild-plugin-manifest')
 const esbuild = require('esbuild')
-const path = require('path')
 const { glob } = require('glob')
 
-const buildAdditionalAssets = buildConfig =>
-  esbuild.build({
+/**
+ * Copy additional assets into distribution
+ * @type {BuildStep}
+ */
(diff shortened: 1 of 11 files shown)


esbuild/assets.config.js
```diff
@@ -31,6 +32,12 @@ const buildAssets = buildConfig => {
       clean({
         patterns: glob.sync(buildConfig.assets.clear),
       }),
+      manifestPlugin({
+        generate: entries =>
+          Object.fromEntries(
+            Object.entries(entries).map(([key, value]) => [key.replace('dist', ''), value.replace('dist', '')]),
+          ),
```
ushkarev, line 39: i have a sneaking suspicion that [options.shortNames](https://github.com/jfortunato/esbuild-plugin-manifest?tab=readme-ov-file#optionsshortnames) might achieve the same result, but more clearly. did you try it?

ushkarev: i take that back. i tried all combos of `shortNames` and they either strip too much or not enough.

so you might as well keep this `generate` rewrite instead. a potential suggestion: "dist" would be replaced everywhere in the path so could feasibly remove something unexpected. maybe use `/^dist\//` instead??

ushkarev (approved):

psoleckimoj (approved):
LGTM!

PR #378, opened 2024-06-19
Add dotenv to ESBuild server watch

Thomas-Geraghty:
- Fixed issue where .env files are not preloaded when using the new ESBuild watch setup


psoleckimoj (approved):
LGTM!

PR #384, opened 2024-07-02
Change sqs assertion

ReedSoftware:
Change sqs assertion to be more functional and not be flaky for slower computers


petergphillips (approved):
LGTM

PR #401, opened 2024-08-06
Update husky

thomasridd:
Quick fix to remove lines from husky deprecated in v9.0.1

see https://github.com/typicode/husky/releases/tag/v9.0.1


andrewrlee (approved):
LGTM

PR #414, opened 2024-09-05
Supporting starting a fresh application

andrewrlee:
I've added a set of client credentials into the hmpps-auth seed database that can be used by anyone running docker locally.

* Fixed an issue where we can't create sessions when running in docker compose due to https only flag.
* Adding example.env file to allow running locally using npm.
* Fix issue with missing redis hostname in docker-compare
* Improve docs

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,17 +28,19 @@ services:
     environment:
       - PRODUCT_ID=UNASSIGNED
       - REDIS_ENABLED=false
+      - REDIS_HOST=localhost
       - HMPPS_AUTH_EXTERNAL_URL=http://localhost:9090/auth
       - HMPPS_AUTH_URL=http://hmpps-auth:8080/auth
       # These will need to match new creds in the seed auth service auth
-      - AUTH_CODE_CLIENT_ID=client-id
-      - AUTH_CODE_CLIENT_SECRET=client-secret
-      - CLIENT_CREDS_CLIENT_ID=system-client-id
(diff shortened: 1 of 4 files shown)


server/config.ts
```diff
@@ -54,7 +54,7 @@ export default {
   gitRef: get('GIT_REF', 'xxxxxxxxxxxxxxxxxxx', requiredInProduction),
   branchName: get('GIT_BRANCH', 'xxxxxxxxxxxxxxxxxxx', requiredInProduction),
   production,
-  https: production,
+  https: process.env.NO_HTTPS === 'true' ? false : production,
```
andrewrlee, line 57: This was preventing log-in as it couldn't create the session cookie before redirecting to auth. 
When Auth redirected back the application refused to accept the request as it couldn't validate state.

docker-compose.yml
```diff
@@ -28,17 +28,19 @@ services:
     environment:
       - PRODUCT_ID=UNASSIGNED
       - REDIS_ENABLED=false
+      - REDIS_HOST=localhost
```
andrewrlee, line 31: this is required when production == true

ushkarev (approved):

docker-compose.yml
Thomas-Geraghty: YAY! No longer have to have a separate changeless on the template just to keep this working locally 😄

server/config.ts
andrewrlee: it was a right pain to track down 😅

PR #416, opened 2024-09-06
Load `*.env` files using only `dotenv`

ushkarev:
There are a few benefits of doing this:
• there is only one tool that loads all `*.env` files so there cannot be inconsistencies between how it's read. e.g. you could not have included `# comment` lines when loading via `export $(cat file.env)`, now you can include comments
• the `start-feature:dev` script (but not `start-feature`) was in fact loading `.env` _on top of_ `feature.env` which could create confusing problems


psoleckimoj (approved):
LGTM!

PR #427, opened 2024-09-10
SDIT-2037: ✨ Latest versions of hmpps orb etc.

petergphillips:


.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ parameters:
 
   node-version:
     type: string
-    default: 20.12-browsers
+    default: 20.17-browsers
```
petergphillips, line 21: this was a bit odd - the `Dockerfile` is running against `20.17` but we're testing against `20.12` 😱.  Guess that's what happens when you let renovate do the upgrades 🤷

psoleckimoj (approved):
LGTM!

PR #439, opened 2024-09-25
Remove 'uuid' dependency and replace with crypto.randomUUID()

tpmcgowan:
Use Node's `crypto.randomUUID()` instead of requiring the `uuid` module - there's no no need if just using v4 UUIDs (see https://www.npmjs.com/package/uuid).


psoleckimoj (approved):
LGTM!

PR #441, opened 2024-09-25
Removing dependency on dotenv

andrewrlee:
Removing dependency on [dotenv](https://www.npmjs.com/package/dotenv). 

Using Node's `-env-file` [mechanism](https://nodejs.org/dist/latest-v20.x/docs/api/cli.html#--env-fileconfig) instead of requiring the dotenv module.


psoleckimoj (dismissed):
LGTM!

esbuild/esbuild.config.js
```diff
@@ -62,7 +62,7 @@ const main = () => {
     let serverProcess = null
     chokidar.watch(['dist']).on('all', () => {
       if (serverProcess) serverProcess.kill()
-      serverProcess = spawn('node', ['-r', 'dotenv/config', 'dist/server.js'], { stdio: 'inherit' })
+      serverProcess = spawn('node', ['--env-file=.env', 'dist/server.js'], { stdio: 'inherit' })
```
psoleckimoj, line 65: Won't this be different for int tests? Not sure how it all relates to esbuild.

psoleckimoj (approved):
LGTM!

esbuild/esbuild.config.js
andrewrlee: yes, it will - added a separate flag for the test server..

package.json
```diff
@@ -10,8 +10,8 @@
     "build": "node esbuild/esbuild.config.js --build",
     "start": "node $NODE_OPTIONS dist/server.js | bunyan -o short",
     "start:dev": "concurrently -k -p \"[{name}]\" -n \"ESBuild,Node\" -c \"yellow.bold,cyan.bold\" \"node esbuild/esbuild.config.js --build --watch\" \"node esbuild/esbuild.config.js --dev-server\"",
-    "start-feature": "DOTENV_CONFIG_PATH=feature.env node $NODE_DEBUG_OPTION -r dotenv/config dist/server.js | bunyan -o short",
-    "start-feature:dev": "DOTENV_CONFIG_PATH=feature.env concurrently -k -p \"[{name}]\" -n \"ESBuild,Node\" -c \"yellow.bold,cyan.bold\" \"node esbuild/esbuild.config.js --build --watch\" \"node esbuild/esbuild.config.js --dev-server\"",
+    "start-feature": " node $NODE_DEBUG_OPTION --env-file=feature.env dist/server.js | bunyan -o short",
+    "start-feature:dev": "concurrently -k -p \"[{name}]\" -n \"ESBuild,Node\" -c \"yellow.bold,cyan.bold\" \"node --build --watch\" \"node esbuild/esbuild.config.js --dev-test-server\"",
```
ushkarev, line 14: this won't work since you removed `esbuild/esbuild.config.js`

psoleckimoj (approved):
LGTM!

package.json
andrewrlee: Thanks @ushkarev I've fixed that now

ushkarev (approved):

xoen:
FYI: Something to bear in mind is that someone in our team was seeing some weirdness just after switching to Node's `--env-file`: A previously perfectly working `.env` file is now not setting environment variables correctly anymore.

It was very confusing especially as Node didn't spit out any error at all...

...turns out that `dotenv` accepted lines with colons, e.g. `KEY: value` while Node's `--env-file` _**silently**_ ignores them (I wish Node errored out instead).

Not a biggie but I'm writing this here in case someone else is stumbling into the same issue.

TL;DR: If you have problems check that `.env` files only have lines like `KEY=value`

CC: @jonesj2193

psoleckimoj:
`.env` files are always `X=Y` format, the dotenv modules also doesn't mention supporting `A: B` since that wouldn't be an env file.

xoen:
Agreed. And yet somehow it worked before 😅 - Just to be clear here: I 100% think Node's ignoring `KEY: value` in env files is the right thing to do...I also found it bizarre that dotenv seemed to somehow be ok with it.

But I thought I leave a comment here in case someone else somehow had a file with colons and can't explain why it stopped working as before.

ushkarev:
> .env files are always X=Y format 

evidently not! different tools consider various quotes and comments in different ways

xoen:
> Just to be clear here: I 100% think Node's ignoring KEY: value in env files is the right thing to do

Actually having said that, my preference would have been for Node to throw some useful error pointing to this, but we can't have everything :P

xoen:
BTW I think this PR is great. It's great that we can replace a dependency with a functionality built into Node itself ❤️ (in case my comment sounded like a criticism to this, which it was not ☺️)

PR #461, opened 2024-10-16
Heat 353 - Typescript Github workflows for test -> build -> deploy

james-jdgtl:
This is the initial release for deploying applications using Github Actions/ghcr.io rather than CircleCI/quay.io

The workflow has been validated on test projects - this is a prototype for Typescript deployments.


.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,104 @@
...
+  deploy_dev:
+    name: Deploy 
+    needs: 
+      - build
+      - helm_lint
+    uses: ministryofjustice/hmpps-github-actions/.github/workflows/deploy_env.yml@v1 # WORKFLOW_VERSION
+    secrets: inherit
+    with:
+      environment: 'dev'
+      app_version: '${{ needs.build.outputs.app_version }}'
```
petergphillips, line 104: don't understand how this works - the workflow dispatch calls it `version` but can't be using that?

petergphillips (approved):
LGTM

PR #463, opened 2024-10-22
HEAT-382: Move appInsights config to JSON config

psoleckimoj:


Dockerfile
```diff
@@ -52,6 +52,7 @@ FROM base
 COPY --from=build --chown=appuser:appgroup \
         /app/package.json \
         /app/package-lock.json \
+        /app/package-lock.json \
```
ushkarev, line 55: this is identical to the line above

PR #470, opened 2024-10-28
Move to eslint 9 using HMPPS eslint rules package

psoleckimoj:


server/routes/testutils/appSetup.ts
```diff
@@ -1,5 +1,5 @@
 import express, { Express } from 'express'
-import { NotFound } from 'http-errors'
```
ushkarev, line 2: I don’t understand why typescript was happy with what we had but eslint is not

ushkarev: Actually, [this](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/named.md) suggests that the rule is supposed to be disabled in the typescript config. This reinforces the point that our eslint rules package is incorrectly mixing eslint v9 and legacy plugins as described in [PR description here](https://github.com/ministryofjustice/hmpps-typescript-lib/pull/11)

ushkarev: In one of our projects, I decided to turn off the `import/named` rule if that’s what their documentation suggests…
https://github.com/ministryofjustice/hmpps-incident-reporting/pull/255/commits/b121c7667809b0d12d275134783f3fde44383262#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2

andrewrlee: Think it makes sense to disable this rule in our eslint config - sounds like it would be checked by the compiler anyway

ushkarev: or rather, we should use the proper eslint v9 typescript plugins!

Thomas-Geraghty (approved):

andrewrlee (approved):
LGTM

PR #474, opened 2024-10-29
Moving to Node 22

andrewrlee:


petergphillips (approved):
LGTM

tsconfig.json
```diff
@@ -1,15 +1,13 @@
 {
+  "extends": "@tsconfig/node22/tsconfig.json",
```
andrewrlee, line 2: Now uses these defaults:
```
    "lib": ["es2023"],
    "module": "node16",
    "target": "es2022",
```

ushkarev (approved):

PR #481, opened 2024-11-15
feat: replace csurf with csrf-sync

ReedSoftware:


server/middleware/setUpCsrf.ts
```diff
@@ -8,7 +8,17 @@ export default function setUpCsrf(): Router {
...
   // CSRF protection
   if (!testMode) {
-    router.use(csurf())
+    const {
+      csrfSynchronisedProtection, // This is the default CSRF protection middleware.
+    } = csrfSync({
+      // By default, csrf-sync uses x-csrf-token header, but we use the token in forms and send it in the request body, so change getTokenFromRequest so it grabs from there
+      getTokenFromRequest: req => {
+        // eslint-disable-next-line no-underscore-dangle
+        return req.body._csrf
```
Thomas-Geraghty, line 17: Is there any issue with us moving to using the header version? Seems like having in it our forms makes it easier to mess up.

ushkarev: but isn't the whole purpose to have it in the form as a hidden field? otherwise wouldn't we have to use ajax or whatnot to add a special header to all form submissions?

Thomas-Geraghty: Ah yes, my brain evidently isn't turned on today. I was thinking about cookies!

andrewrlee (approved):
LGTM

PR #489, opened 2024-12-04
Prevent false positives when creating new services

andrewrlee:
We always get a security alert raised for every new service due to github identifying the base64 encoded version of the credentials as a high entropy string. 
This removes that hardcode string and calculates it on the fly


psoleckimoj (approved):
LGTM!

server/authentication/clientCredentials.test.ts
```diff
@@ -2,7 +2,8 @@ import generateOauthClientToken from './clientCredentials'
 
 describe('generateOauthClientToken', () => {
   it('Token can be generated', () => {
-    expect(generateOauthClientToken('bob', 'password1')).toBe('Basic Ym9iOnBhc3N3b3JkMQ==')
+    const base64Creds = Buffer.from('bob:secret').toString('base64')
+    expect(generateOauthClientToken('bob', 'secret')).toBe(`Basic ${base64Creds}`)
```
ushkarev, line 6: i don't hugely care but this test is basically now doing little but asserting that 2 calls to `base64('bob:secret')` return the same thing 🤣

sure, high entropy _can_ be suspiciously like a password, but is that an argument to never hard code test strings?

PR #511, opened 2025-02-10
Update ts config file to remove duplicated config

davidthomas22:
@andrewrlee as discussed remove duplicated config from tsconfig file.

https://www.npmjs.com/package/@tsconfig/node22


andrewrlee (approved):
LGTM

PR #518, opened 2025-03-04
Move to using Alpine as a base image.

alexw-moj:
PR for discussion about moving to Alpine as the base image.

Size comparison when built:

| Image                    | Size   |
|--------------------------|--------|
| node:22-alpine           | 220 MB |
| node:22.14-bookworm-slim | 307 MB |

`version` has been removed from the docker-compose files as it has been [deprecated](https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-top-level-element-obsolete).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,10 @@ FROM node:22-alpine AS base
 
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
-RUN apk add --no-cache tzdata
+RUN apk --update-cache upgrade --available \
+  && apk --no-cache add tzdata \
+  && rm -rf /var/cache/apk/*
+
 ENV TZ=Europe/London
 RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezone
 


Dockerfile
```diff
@@ -1,20 +1,21 @@
...
-        adduser --uid 2000 --system appuser --gid 2000
+    adduser --uid 2000 --system appuser --ingroup appgroup
 
 WORKDIR /app
 
+ARG BUILD_NUMBER
+ARG GIT_REF
+ARG GIT_BRANCH
+
 # Cache breaking and ensure required build / git args defined
```
Thomas-Geraghty, line 19: Not necessarily related to this PR, but maybe we need to update this line, as I believe the build number is no longer used for cache-busting as of https://github.com/ministryofjustice/hmpps-template-typescript/pull/377

Thomas-Geraghty:
Looks good to me, but I've used this on existing projects, so would be good to get someone else's review @ministryofjustice/hmpps-typescript

psoleckimoj (approved):
LGTM!

Dockerfile
```diff
@@ -25,13 +26,8 @@ ENV BUILD_NUMBER=${BUILD_NUMBER}
 ENV GIT_REF=${GIT_REF}
 ENV GIT_BRANCH=${GIT_BRANCH}
 
-RUN apt-get update && \
```
andrewrlee, line 28: This upgraded any local dependencies to include security fixes, is there an alpine equivalent?

alexw-moj: Happy to add that. As he says in the linked comment `upgrading the image used for the build is not particularly useful. Upgrading the deployment base image makes much more sense.`

andrewrlee: yes - I think previously it would update the base which the deployment image inherited from which is aligned with that?

alexw-moj: Updating added in https://github.com/ministryofjustice/hmpps-template-typescript/pull/518/commits/97d2f104ef892189d7e9ce456f36cf7be3f72083

andrewrlee (approved):
LGTM

PR #521, opened 2025-03-07
Fix issue with AppInsights and multiple paths

andrewrlee:
Requests for routes that define multiple patterns were failing to be processed by appinsights and were not being exported e.g:

```ts
get(['/', '/:monitorType/:monitorName'], async (req, res) => {..
```

This would coerce the array into a string and fail with an error from app insights complaining about not accepting `,`s

This now reports these kind of paths as an operation with a name like:

`GET "/" | "/:monitorType/:monitorName"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Change log
 
+**March 7th 2025** - Fix issue with AppInsights
+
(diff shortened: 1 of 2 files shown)


psoleckimoj (approved):
LGTM!

PR #522, opened 2025-03-10
Add example-api, using `@ministryofjustice/hmpps-rest-client`

Thomas-Geraghty:


server/data/exampleApiClient.ts
```diff
@@ -0,0 +1,41 @@
...
+  getSystemClientTokenFromHmppsAuth,
+  asSystem,
+} from '@ministryofjustice/hmpps-rest-client'
+import config from '../config'
+import logger from '../../logger'
+
+export default class ExampleApiClient extends RestClient {
+  constructor(tokenStore: TokenStore) {
+    super('Example API', config.apis.exampleApi, logger, (username: string) =>
+      getSystemClientTokenFromHmppsAuth(config.apis.hmppsAuth, tokenStore, logger, username),
```
andrewrlee, line 13: With hindsight, it might be better to wrap this in a class and explicitly encourage it to be reused. 

```typescript
const hmppsAuthTokenClient = new HmppsAuthTokenClient(config.apis.hmppsAuth, tokenStore, logger)
const exampleApiClient = new ExampleApiClient(hmppsAuthTokenClient)
```
which would then change this to: 
```typescript
constructor(hmppsAuthTokenClient: HmppsAuthTokenClient) {
    super('Example API', config.apis.exampleApi, logger, hmppsAuthTokenClient)
}
```
...
```typescript
 const token = hmppsAuthTokenClient.getToken(username)
```

Thomas-Geraghty: Cool with me - I think I'll make a `AuthenticationClient` interface or something that needs to implement a `getToken` endpoint, then I'll build a HMPPS Auth one and include that in the template

Thomas-Geraghty: Actioned in https://github.com/ministryofjustice/hmpps-template-typescript/pull/522/commits/3079913518dd718504ff4c3318bc587da8b32748

Thomas-Geraghty: On another note, breaking this into the `HmppsAuthTokenClient` has given a quite clear option for breaking this out into it's own `hmpps-typescript-lib` module

server/data/exampleApiClient.ts
```diff
@@ -0,0 +1,34 @@
...
+   *   getCurrentTime(token: string) {
+   *     return this.get({ path: '/example/time' }, asUser(token))
+   *   }
+   */
+
+  /**
+   * If wanting to make a call using a System token,
+   * tied to a User (i.e. for backend auditing purposes)
+   *
+   * import { asUser } from '@ministryofjustice/hmpps-rest-client'
```
andrewrlee, line 28: little typo.. `asSystem`

Thomas-Geraghty: Totally rewritten in https://github.com/ministryofjustice/hmpps-template-typescript/pull/522/commits/fd8e10d61a3d8123bbe9007a92b0dc18f1557e0f - these were just drafts and I forgot to rewrite!

server/data/tokenVerificationClient.ts
```diff
@@ -0,0 +1,39 @@
...
+          request.verified = true
+          return true
+        }
+
+        return false
+      })
+      .catch((e) => {
+        logger.debug(`Token verification failed for user "${user.username}'`)
+        return false
+      })
```
andrewrlee, line 37: is there a reason you're rewriting this to not use async await?

Thomas-Geraghty: Personal preference thing - I'm not a really big fan of async/await because of the try/catch block that you end up needing to wrap stuff with. 

This was initially missing that (though I think it was safe at the time because the error was caught in the function itself, and instead nothing was returned). So I just defaulted back to my normal. Happy to switch it to async/await (though I'll do that when moving it to it's own package as discussed on Slack)

andrewrlee: I think GDS [generally recommended](https://gds-way.digital.cabinet-office.gov.uk/manuals/programming-languages/nodejs/#use-of-async-await) using async await

psoleckimoj (approved):
LGTM!

PR #525, opened 2025-03-12
Replace auth/tokenVerifier logic with `hmpps-auth-clients`

Thomas-Geraghty:
Replace/cleanup large amounts of the existing logic/boilerplate for HMPPS Auth and Token Verification API with the new `hmpps-auth-clients` library


Thomas-Geraghty:
Putting back into draft until we've managed to test it on a few projects :)

andrewrlee (approved):
LGTM

PR #527, opened 2025-03-19
Only restart dev server once esbuild complete

froddd:
## Context

In the context of the esbuild routine, the current approach to watching for file changes means that, when running a development server, a new server process is created and killed for _every single file found in the `dist` directory_. I have [replicated this with useful logging on a branch of our own repository](https://github.com/ministryofjustice/hmpps-approved-premises-ui/tree/esbuild-proceses-spawn-count), for which we've recently back-ported the template's esbuild config: when running the `npm run start:dev` command, you will see hundreds of `Started new server process with pid ...` being logged to the console.

This isn't an issue in itself (as long as we're happy that starting/killing processes willy-nilly is fine 😄), but, should you decide to keep your server running, while also attempting to run the integration tests in UI mode (`npm run test:integration:ui` in the linked demo of the issue, something we do regularly on our repository), a `pgrep` error occurs, killing both the integration test and server processes:

```
node:events:496
      throw er; // Unhandled 'error' event
      ^

Error: spawn pgrep EAGAIN
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  errno: -35,
  code: 'EAGAIN',
  syscall: 'spawn pgrep',
  path: 'pgrep',
  spawnargs: [ '-P', 48347 ]
}
```

A little bit of digging seemed to point to this pgrep error being linked to too many node processes/high PIDs, or something along those lines.

## Solution

Introduce a debounce on all calls to restart the server or build the app or assets (as suggested by @Thomas-Geraghty 😉). The debounce delay is set to 100ms, which seems sufficient to not hinder speed but also to prevent the issue described above.

### Also 

A couple of `process.stderr` calls have been switched to `process.stdout`, as they were not used to display errors.

### DIFF
--- a/esbuild/esbuild.config.js
+++ b/esbuild/esbuild.config.js
@@ -9,6 +9,20 @@ const buildApp = require('./app.config')
 const cwd = process.cwd()
 
 /**
+ * Simple debounce helper
+ * @param {Function} fn - The function to debounce
+ * @param {number} delay - The delay in ms
+ * @returns {Function}
+ */
+function debounce(fn, delay = 200) {
+  let timeout
+  return (...args) => {
+    clearTimeout(timeout)
+    timeout = setTimeout(() => fn(...args), delay)
+  }
+}
+
+/**
  * Configuration for build steps
  * @type {BuildConfig}
  */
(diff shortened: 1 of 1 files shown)


Thomas-Geraghty (dismissed):

esbuild/app.config.js
```diff
@@ -30,7 +30,7 @@ const buildApp = buildConfig => {
  * @returns {Promise}
  */
 module.exports = buildConfig => {
-  process.stderr.write('\u{1b}[1m\u{2728} Building app...\u{1b}[0m\n')
+  process.stdout.write('\u{1b}[1m\u{2728} Building app...\u{1b}[0m\n')
```
ushkarev, line 33: no big deal really but logging normally goes to stderr

ushkarev (approved):
how come you wanted to swap logging to stdout?

esbuild/app.config.js
froddd: Messages logged to `stderr` don't automatically get redirected to output, but shown to the user instead. This is logging output for the script, so should probably be redirected to output rather than treated as error messaging. 

In practice, if I were to pipe running this script into a log file, using `stderr` would add nothing to the log file  (unless explicitly combining error with standard output with `2>&1`). Using `stdout` means the log would contain each of the steps of the build.

No biggie really as you say, and I doubt there's much use case for actually saving logs of the watched build! It just felt like an oversight/typo, as this isn't actually an error message being output. Happy to change back to `stderr` if we insist on this as standard.

ushkarev: i don't insist for logging to go back to stderr!
but i'd actually say that if i saw logging going to stdout that _that_ would look like an oversight to me 🤣

froddd: Fair enough -- I've removed the change to go back to logging to `stderr` by default.

ushkarev (approved):

esbuild/app.config.js
ushkarev: 🙈 i really wasn't demanding that change! 🙈
but yay, i am hoping that the debounce thingy helps in one of our projects which re/starts really really slowly locally in dev mode

PR #540, opened 2025-04-08
Fixing error handling

andrewrlee:


server/routes/index.ts
```diff
@@ -13,15 +11,8 @@ export default function routes({ auditService, exampleService }: Services): Rout
   get('/', async (req, res, next) => {
     await auditService.logPageView(Page.EXAMPLE_PAGE, { who: res.locals.user.username, correlationId: req.id })
 
-    try {
-      const currentTime = await exampleService.getCurrentTime()
-
-      return res.render('pages/index', { currentTime })
-    } catch (e) {
-      logger.error(e)
```
andrewrlee, line 21: We should encourage inheriting teams to just allow errors to propagate from controllers and be handled by the global error handler. 
Ideally we'd only log error events once and not leak data layer implementation details to the web layer.

server/services/exampleService.test.ts
```diff
@@ -1,22 +1,20 @@
 import ExampleApiClient from '../data/exampleApiClient'
 import ExampleService from './exampleService'
 
+jest.mock('../data/exampleApiClient')
```
andrewrlee, line 4: bit neater using jest.mocks

psoleckimoj (dismissed):
LGTM!

psoleckimoj (approved):
LGTM!

server/routes/index.ts
Thomas-Geraghty: It would be good to get an example on how to do this correctly. I work with these day in, day out, yet even I'm not aware of what the intended solution is - especially considering what is being suggested here can lead to things like sessions getting terminated due to 401 propagation.

andrewrlee: This has all been documented [here now](https://github.com/ministryofjustice/hmpps-typescript-lib/blob/main/packages/rest-client/README.md#error-handling)

PR #560, opened 2025-06-24
Upgrade to express 5

psoleckimoj:


server/routes/testutils/appSetup.ts
```diff
@@ -38,6 +38,7 @@ function appSetup(services: Services, production: boolean, userSupplier: () => H
     res.locals = {
       user: { ...req.user } as HmppsUser,
     }
+    req.body = req.body || {}
```
psoleckimoj, line 41: Can be `undefined` in express 5.

psoleckimoj: This stops it defaulting to `undefined` ?

ushkarev: #561 removes it and all tests are perfectly happy still!

PR #562, opened 2025-06-26
Add user data to appInsights requests

psoleckimoj:
Pulled in from Visits.... https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/main/server/utils/azureAppInsights.ts

Allows for easier debugging/diagnosis of faults, data breaches etc

### DIFF
--- a/server/utils/azureAppInsights.test.ts
+++ b/server/utils/azureAppInsights.test.ts
@@ -1,0 +1,82 @@
+import { DataTelemetry, EnvelopeTelemetry } from 'applicationinsights/out/Declarations/Contracts'
+import { addUserDataToRequests, ContextObject } from './azureAppInsights'
+
+const user = {
+  activeCaseLoadId: 'LII',
+  username: 'test-user',
+}
+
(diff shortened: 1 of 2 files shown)


ushkarev (approved):
is this relevant to probation? as i understand it, the template should work for both prisons and probation.

since we’re copying stuff from other repos already, i had a look at our team’s ones and we have a little more in here to ignore boring paths:

https://github.com/ministryofjustice/hmpps-non-associations/blob/main/server/utils/azureAppInsights.ts#L49-L61 (and corresponding test)

worth adding?

looking at the history, this itself was derived from BASM. i don’t remember the conversations from the time, i bet it was Peter’s original idea :)

psoleckimoj:
@ushkarev it was @rj-adams that requested it to allow easier debugging/fault finding etc. Definitely worth adding in 🙂 I pulled this from https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/main/server/utils/azureAppInsights.ts

PR #575, opened 2025-08-07
fix restarting

carloveo-moj:
upgrading chokidar to v4 meant the paths needed changing.

See https://www.npmjs.com/package/chokidar#upgrading for details


andrewrlee (approved):
LGTM

PR #582, opened 2025-08-27
Split out jest/lint-staged configs

Thomas-Geraghty:
- Split out jest config
- Split out lint-staged config
  - Added `.mjs` to lint-staged as we now have multiple .mjs files in the repo
- Cleaned up package.json


lint-staged.config.mjs
```diff
@@ -0,0 +1,4 @@
+export default {
+  '*.{ts,js,mjs,css}': ['prettier --write', 'eslint --fix'],
```
ushkarev, line 2: i note that you’ve added mjs so made me think whether we want all of `[c|m]?js` and `[c|m]?ts`

PR #584, opened 2025-09-04
Moving from cypress to playwright

andrewrlee:
WIP:
* Need to add custom GHA for running integration tests
* Adjust shared eslint config to accept playwright defaults


integration_tests/pages/homePage.ts
```diff
@@ -0,0 +1,17 @@
...
+    super(page)
+    this.header = page.locator('h1', { hasText: 'This site is under construction...' })
+  }
+
+  static async verifyOnPage(page: Page): Promise<HomePage> {
+    const homePage = new HomePage(page)
+    await expect(homePage.header).toBeVisible()
+    return homePage
+  }
+}
```
andrewrlee, line 17: we could have had something generic on `AbstractPage` like we did in cypress but it's a bit unwieldy:
```ts
static verifyOnPage<T>(constructor: new (page: Page) => T, page: Page): T {
    const result = new constructor(page)
    expect(result.locationVerifier).isVisible()
    return result
}
...
 class MyPage {
   constructor(page:Page) {
     super(page, page.locator(...))
   }
 }
...
const myPage = AbstractPage.verifyOnPage(MyPage, page)
```

server/views/pages/index.njk
```diff
@@ -7,6 +7,6 @@
 
   <h1>This site is under construction...</h1>
   <p>Please check back later when there is content to view.</p>
-  <p id="timestamp">The time is currently {{ currentTime }}</p>
+  <p data-qa="timestamp">The time is currently {{ currentTime }}</p>
```
andrewrlee, line 10: using standard data-qa attribute

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+  outputDir: './test_results/playwright/test-output',
+  testDir: './integration_tests/specs',
+  /* Maximum time one test can run for. (millis) */
+  timeout: 3 * 60 * 1000,
+  /* Maximum time test suite can run for. (millis) */
+  globalTimeout: 60 * 60 * 1000,
+  /* Run tests in files in parallel */
+  fullyParallel: false,
+  /* Ensure tests run consecutively due to inability to share wiremock instance */
+  workers: 1,
```
andrewrlee, line 23: needs to run strictly in parallel or wiremock gets confused

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 1 : 0,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: [
+    ['list'],
+    ['html', { outputFolder: 'test_results/playwright/report', open: process.env.CI ? 'never' : 'on-failure' }],
+    ['junit', { outputFile: 'test_results/playwright/junit.xml' }],
```
andrewrlee, line 32: this keeps things nice and tidy
```
test_results
├── jest
│   └── junit.xml
├── playwright
│   ├── junit.xml
```

playwright.config.ts
Thomas-Geraghty: I wonder if we can get a solution for this going, not running in parallel feels like it's skipping a chunk of the benefit we get from switching to Playwright

package.json
```diff
@@ -36,6 +37,7 @@
     "@ministryofjustice/hmpps-auth-clients": "^0.0.3",
     "@ministryofjustice/hmpps-monitoring": "^0.0.2",
     "@ministryofjustice/hmpps-rest-client": "^0.0.3",
+    "@playwright/test": "^1.56.1",
```
andrewrlee: It also has 90 fewer dependencies that cypress too: 1137 > 1047

PR #598, opened 2025-10-16
Vite POC

andrewrlee:
Seems faster and reduces the number of dependencies from 1219 ->  998


server/data/exampleApiClient.test.ts
```diff
@@ -1,23 +1,24 @@
+import { beforeEach, afterEach, Mocked, vi, expect, it, describe } from 'vitest'
```
andrewrlee: By default vite requires explicit import but you can configure it to use globals which might make a migration more easy.. I hit linting/tsc issues which probably can be overcome.

andrewrlee:
Would need to update the GHA:
```
Reading archive npm_unit_test_artifacts.zip
  Report name: Test Report
  Skipping vitest/junit.xml: filename does not match pattern
  Skipping vitest/unit-test-report/assets/index-xxxxxxxxxx.js: filename does not match pattern
  Skipping vitest/unit-test-report/assets/index-xxxxxxxxxx.css: filename does not match pattern
  Skipping vitest/unit-test-report/bg.png: filename does not match pattern
  Skipping vitest/unit-test-report/favicon.ico: filename does not match pattern
  Skipping vitest/unit-test-report/favicon.svg: filename does not match pattern
  Skipping vitest/unit-test-report/html.meta.json.gz: filename does not match pattern
  Skipping vitest/unit-test-report/index.html: filename does not match pattern
Creating test report Test Report
  Warning: No file matches path jest/**/*.xml
```

Artifacts gets uploaded up but require running with vite preview to view:
```
 HTML  Report is generated
       You can run npx vite preview --outDir test_results/vitest/unit-test-report to see the test results.
JUNIT report written to /home/runner/work/hmpps-template-typescript/hmpps-template-typescript/test_results/vitest/junit.xml
```

PR #667, opened 2026-02-05
Alee/update to latest deps

andrewrlee:


psoleckimoj:
@andrewrlee is this still valid?

PR #668, opened 2026-02-05
Switch to using `@ministryofjustice/hmpps-azure-telemetry` for Azure telemetry

Thomas-Geraghty:
- Removed old telemetry
- Added `@ministryofjustice/hmpps-azure-telemetry` package and set it up with some good filtering defaults


server.ts
```diff
@@ -1,5 +1,4 @@
-// Require app insights before anything else to allow for instrumentation of bunyan and express
```
andrewrlee, line 1: is it worth keeping in the comment? or does auto instrumentation not happen in the same way?

andrewrlee (approved):
LGTM

PR #717, opened 2026-04-14
Update client-side typescript config for v6

ushkarev:
Continues on from #715

### DIFF
--- a/assets/js/tsconfig.json
+++ b/assets/js/tsconfig.json
@@ -1,14 +1,14 @@
 {
   "$schema": "https://www.schemastore.org/tsconfig",
   "compilerOptions": {
-    "module": "commonjs",
+    "strict": false,
     "target": "es2020",
     "noEmit": true,
     "lib": ["dom", "dom.iterable", "es2020"],
     "esModuleInterop": true,
     "skipLibCheck": true,
     "sourceMap": true,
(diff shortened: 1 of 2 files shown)


assets/js/tsconfig.json
```diff
@@ -1,14 +1,14 @@
 {
   "$schema": "https://www.schemastore.org/tsconfig",
   "compilerOptions": {
-    "module": "commonjs",
+    "strict": false,
```
ushkarev, line 4: Leaving strict turned off by default like root settings otherwise the client-side would be more strict than the server…

Setting module is unnecessary (I think?) since `tsc` doesn't emit anything and seems best left off. Plus, module & moduleResolution are mysterious settings that are hard to get right and some combos are now deprecated.

package.json
```diff
@@ -14,7 +14,7 @@
     "start-feature:dev": "node esbuild/esbuild --watch --env feature.env",
     "lint": "eslint . --cache --max-warnings 0",
     "lint-fix": "eslint . --cache --max-warnings 0 --fix",
-    "typecheck": "tsc && tsc -p integration_tests",
+    "typecheck": "tsc && tsc -p integration_tests && tsc -p assets/js",
```
ushkarev, line 17: We’d want to check client-side code too, right?

psoleckimoj (approved):
lgtm!

psoleckimoj (approved):
lgtm

PR #718, opened 2026-04-20
Improve the types to be compatible with strict type rules

bill-moj:
Updated type definitions, such that typecheck will pass even if we update `tsconfig.json` with:

```
    "strict": true,
    "noUnusedParameters": true,
```


server/@types/express/index.d.ts
```diff
@@ -23,6 +23,15 @@ export declare global {
 
     interface Locals {
       user: HmppsUser
+      cspNonce: string
+      csrfToken: ReturnType<CsrfTokenGenerator>
```
andrewrlee: would this need to be imported?

bill-moj: i copied this type definition from csrf-sync library. just checked that the return type is actually just a `string` so there isn't really any need to complicate this.

andrewrlee (dismissed):
LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change log
 
+**April 22nd 2026** - Improve type safety
```
andrewrlee:
```suggestion
**April 23rd 2026** - Improve type safety
```

psoleckimoj (approved):
lgtm!

PR #738, opened 2026-05-07
Run docker build on feature branches without publishing

KristianGregor:
## Problem

The `build` job was gated behind `if: github.ref == 'refs/heads/main'`, which means a broken `Dockerfile` can pass all PR checks undetected and only fail after merging to main when the image is built and published.

## Solution

- Remove the `if: github.ref == 'refs/heads/main'` guard from the `build` job so it runs on all branches
- Make `push`, `generate_sbom`, `attach_sbom_to_registry`, and `sign_image` conditional on the main branch — feature branches build but do not publish
- Add an explicit `if: github.ref == 'refs/heads/main'` to `deploy_dev` to ensure it still only runs after a main-branch build

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -101,11 +101,9 @@ jobs:
       slack_notification: 'false' # if 'true', it needs a valid NONPROD_RELEASES_SLACK_CHANNEL variable to be set
   # deploy_preprod:
   #   name: Deploy to pre-production environment
-  #   if: needs.check_publish.outputs.enabled == 'true'
   #   needs:
   #     - build
   #     - deploy_dev
-  #     - check_publish
   #   uses: ministryofjustice/hmpps-github-actions/.github/workflows/deploy_env.yml@v2 # WORKFLOW_VERSION
   #   secrets: inherit
   #   with:
@@ -114,11 +112,9 @@ jobs:
   #     slack_notification: 'false' # if 'true', it needs a valid NONPROD_RELEASES_SLACK_CHANNEL variable to be set
   # deploy_prod:
   #   name: Deploy to production environment
-  #   if: needs.check_publish.outputs.enabled == 'true'
   #   needs:
   #     - build
   #     - deploy_preprod
-  #     - check_publish
   #   uses: ministryofjustice/hmpps-github-actions/.github/workflows/deploy_env.yml@v2 # WORKFLOW_VERSION
   #   secrets: inherit
   #   with:
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -72,9 +72,9 @@ jobs:
       additional_docker_tag: ${{ inputs.additional_docker_tag }}
       push: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
       docker_multiplatform: false
-      generate_sbom: ${{ github.ref == 'refs/heads/main' }}
-      attach_sbom_to_registry: ${{ github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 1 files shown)


ushkarev (dismissed):

ushkarev (dismissed):

.github/workflows/pipeline.yml
```diff
@@ -71,13 +70,14 @@ jobs:
...
       additional_docker_tag: ${{ inputs.additional_docker_tag }}
-      push: ${{ inputs.push || inputs.push == null }}
+      push: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
       docker_multiplatform: false
-      generate_sbom: true
-      attach_sbom_to_registry: true
-      sign_image: true
+      generate_sbom: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
+      attach_sbom_to_registry: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
+      sign_image: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
```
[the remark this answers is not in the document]
KristianGregor: tried this but in my opinion it adds almost as much code as is replaced, so I don't see it as strictly necessary unless the conditions get even more complex in future.

.github/workflows/pipeline.yml
```diff
@@ -62,7 +62,6 @@ jobs:
       environment: ${{ matrix.environments }}
   build:
     name: Build docker image from hmpps-github-actions
-    if: github.ref == 'refs/heads/main'
     uses: ministryofjustice/hmpps-github-actions/.github/workflows/docker_build.yml@v2 # WORKFLOW_VERSION
     needs:
```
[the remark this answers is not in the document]
KristianGregor: The problem is GitHub Actions permissions are static — you can't make them conditional on branch. To implement this properly you'd need two separate build job definitions (one for branches, one for main), which would mean duplicating ~10 lines of config and adding significant maintenance burden to a template. Given that the elevated permissions are already never used on feature branches (the sign/SBOM jobs are gated on inputs.push inside the reusable workflow), the practical security exposure is minimal.

ushkarev (dismissed):

.github/workflows/pipeline.yml
```diff
@@ -71,13 +70,14 @@ jobs:
...
       docker_multiplatform: false
-      generate_sbom: true
-      attach_sbom_to_registry: true
-      sign_image: true
+      generate_sbom: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
+      attach_sbom_to_registry: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
+      sign_image: ${{ github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null) }}
   deploy_dev:
     name: Deploy to the development environment
+    if: github.ref == 'refs/heads/main' && (inputs.push || inputs.push == null)
```
psoleckimoj: Only concern is that we have no prod env for templates and this would still want to be in the `deploy_prod` step 🤔 Not sure the best way to show that though?

KristianGregor: Good shout, I've made a change so the commented out preprod/prod steps have the check included. this was introducing even more duplication of the condition also so I've pulled that out into a run-once job. 

does that look like it would address the concern?

psoleckimoj: Yeah I spotted the commented stuff out after making the comment so exactly what I'd have done 👍

psoleckimoj (dismissed):
lgtm

andrewrlee (approved):
LGTM

PR #774, opened 2026-06-19
Fancier Wiremock interface

ushkarev:
This will make it easier to write stubs correctly, especially when you want to match query parameters (using `urlPattern` is sensitive to parameter order so is not very effective).

Since most apis will want a /health/ping stub, it’s convenient to have something to reuse and for the health check integration tests to be simpler to update.


andrewrlee (dismissed):
LGTM

andrewrlee:
be worth updating the changelog?

andrewrlee (approved):
LGTM

PR #778, opened 2026-06-25
Remove shutdown logging in azureAppInsights

Thomas-Geraghty:
Logger was inadvertently being loaded before the azure-telemetry model, causing instrumentation to be incorrectly configured. Removing this logging fixes AppTraces reporting in Azure

Credits to @tpmcgowan who spotted the issue and suggested the fix


andrewrlee (approved):
LGTM

PR #799, opened 2026-08-05
Adding example audit middleware

andrewrlee:
Main driver is to provide an area where audit can be tested - this is the "before", the next step would be to move to using the new shared library. 

Not sure whether we'd want to include this by default for new teams to adopt - possibly a bit of a hurdle for teams just starting up to add audit integration. 

Also not sure about e2e testing - would we want to test events are being raised using localstack or is that overkill?

Given these uncertainties -a half way house might to make this a temporary change and revert (at least the helm config) in the near future?


helm_deploy/hmpps-template-typescript/values.yaml
```diff
@@ -1,9 +1,9 @@
 generic-service:
   nameOverride: hmpps-template-typescript
   productId: 'UNASSIGNED' # productId for the product that this belongs too, i.e. DPS001, see README.md for details
-
-  # the IRSA service account name for accessing AWS resources
-  # serviceAccountName: "hmpps-template-typescript"
+  
+  # the IRSA service account name for accessing AWS resources, comment out if you haven't configured IRSA in your namespace yet.
```
andrewrlee, line 5: If we decide to leave this in then it will have a knock on effect on documentation and add another hurdle for teams to start a new service.

We could potentially comment this out as part of the bootstrap?

server/data/hmppsAuditClient.ts
```diff
@@ -7,7 +7,7 @@ export interface AuditEvent {
   subjectId?: string
   subjectType?: string
   correlationId?: string
-  details?: object
+  details?: Record<string, unknown>
```
andrewrlee, line 10: Seems to be a bit of a mismatch between the typescript lib (which doesn't complain and this codebase which does (`{} != Record<string, unknown>`)

This might become more of an issue when I move to test the new library

server/middleware/auditSearchRequest.ts
```diff
@@ -0,0 +1,25 @@
+import type { RequestHandler } from 'express'
+import { Page } from '../services/auditService'
+import { Services } from '../services'
+
+export default function auditSearchRequest({ services, page }: { services: Services; page: Page }): RequestHandler {
```
andrewrlee, line 5: I was umming and arring between adding an example of crn vs prison number but went for search request which is neutral across prison and probation domains..

not sure how realistic it is that a team would have reusable middleware for search queries - guess it is possible. 

I passed searchTerm via query string to ease testing but not great practice!

server/routes/testutils/appSetup.ts
```diff
@@ -48,12 +58,12 @@ function appSetup(services: Services, production: boolean, userSupplier: () => H
     next()
   })
   app.use((req, _res, next) => {
-    req.id = randomUUID()
+    req.id = '4d0fd4da-ecc1-454d-8308-cdee6b8b91f7'
```
andrewrlee, line 60: hardcoded a uuid - could have added a seem to make this injectable but seemed a bit overkill

server/middleware/auditSearchRequest.ts
```diff
@@ -0,0 +1,25 @@
...
+    const { searchTerm } = req.query // search terms should be passed via post body rather than query params, but this is to allow testing
+    if (typeof searchTerm === 'string') {
+      await auditService.logAuditEvent({
+        correlationId: req.id,
+        who: user.username,
+        what: page,
+        subjectType: 'SEARCH_TERM',
+        subjectId: searchTerm,
+        details: { build: applicationInfo.gitRef, userRoles: user.userRoles },
+      })
```
andrewrlee, line 20: prisoner profile logs these details which seem like a reasonable default.

helm_deploy/hmpps-template-typescript/values.yaml
psoleckimoj: Bootstrap sounds OK, maybe we need to update and ask if they want Audit? 🤔 People never read instructions 😉

server/routes/index.ts
```diff
@@ -13,5 +15,13 @@ export default function routes({ auditService, exampleService }: Services): Rout
     return res.render('pages/index', { currentTime })
   })
 
+  router.get(
```
psoleckimoj, line 18: Hmm, thought I'd left a comment, do we need a new route just to demo the audit? Can't it just sit in the home route?

psoleckimoj (approved):

helm_deploy/hmpps-template-typescript/values.yaml
andrewrlee: We've agreed to merge as is temporarily and then remove after so we don't need to modify the bootstrap scripts

server/routes/index.ts
andrewrlee: Adding an example of how to do audit "properly" requires a subject - like either a user, prisoner or offender or search.. so couldn't really use the homepage

PR #806, opened 2026-08-14
Moving to npm 12

andrewrlee:
As part of this we're removing dependency on allowlist scripts


Dockerfile
```diff
@@ -17,7 +17,9 @@ RUN test -n "$GIT_BRANCH" || (echo "GIT_BRANCH not set" && false)
 
 WORKDIR /app
 
-COPY package*.json .allowed-scripts.mjs .npmrc ./
+COPY package*.json .npmrc ./
+
+RUN npm install npm@12.0.2
```
andrewrlee: this is a bit lame - we'd need this until it becomes available in a base image.

package.json
```diff
@@ -23,11 +23,11 @@
     "int-test-init:ci": "npx playwright install",
     "clean": "rm -rf dist test_results",
     "rebuild": "npm run clean && npm run setup && npm run build",
-    "setup": "npm ci && hmpps-npm-script-run-allowlist"
+    "setup": "npm ci"
```
andrewrlee, line 26: we could potentially remove this and go back to using `npm ci` everywhere?

package.json
```diff
@@ -98,5 +97,13 @@
...
     "brace-expansion": "5.0.9",
     "js-yaml": "3.15.1"
+  },
+  "allowScripts": {
+    "@parcel/watcher@2.6.0": true,
+    "dtrace-provider@0.8.8": false,
+    "esbuild@0.28.1": true,
+    "fsevents@2.3.3": true,
+    "fsevents@2.3.2": true,
+    "unrs-resolver@1.12.2": true
```
andrewrlee, line 107: by default it pins the version which seems like the correct behaviour

Dockerfile
```diff
@@ -17,7 +17,8 @@ RUN test -n "$GIT_BRANCH" || (echo "GIT_BRANCH not set" && false)
 
 WORKDIR /app
 
-COPY package*.json .allowed-scripts.mjs .npmrc ./
+RUN npm install -g npm@12.0.2
```
psoleckimoj, line 20: We could ask Charlie if it's worth spinning up an `npm 12` image alongside the current one?

psoleckimoj (approved):

PR #817, opened 2026-09-02
SDIT-4187: 🚨 Enable strict / no unused and fix warnings

petergphillips:


server/routes/index.test.ts
```diff
@@ -9,7 +9,7 @@ import { Page } from '.'
 jest.mock('@ministryofjustice/hmpps-audit-client')
 jest.mock('../services/exampleService')
 
-const auditService = new AuditService(undefined) as jest.Mocked<AuditService>
+const auditService = new AuditService({} as never) as jest.Mocked<AuditService>
```
petergphillips, line 12: would have preferred to do a `{} as HmppsAuditClient` but that type isn't exported

server/routes/testutils/appSetup.ts
```diff
@@ -61,7 +61,7 @@ function appSetup(services: Services, production: boolean, userSupplier: () => H
   })
   app.use(express.json())
   app.use(express.urlencoded({ extended: true }))
-  app.use(routes({ applicationInfo, ...services }))
+  app.use(routes({ applicationInfo, ...services } as Services))
```
petergphillips, line 64: `services` already has an `applicationInfo` so have to play around and trick typescript here

server/routes/index.ts
```diff
@@ -25,7 +25,7 @@ export default function routes(services: Services): Router {
   router.get(
     '/perform-search',
     auditSearchRequest({ services, page: Page.SEARCH_OFFENDERS }),
-    async (req, res, _next) => {
+    async (_req, res, _next) => {
```
petergphillips, line 28: unused

psoleckimoj (approved):

PR #822, opened 2026-09-09
Adding basic user metadata to app insights telemetry

brightonsbox:
This PR adds basic user metadata to app insights telemetry out of the box, including the user UUID stored in HMPPS Auth and, for prison users, the activeCaseLoadId.  This should be considered the minimal set of data to be added to
telemetry for tactical audit and debugging user specific issues.


server/utils/azureAppInsightsUserMetadata.ts
```diff
@@ -0,0 +1,32 @@
...
+    const user = (res.locals?.user ?? {}) as HmppsUser
+
+    telemetry.setSpanAttributes({
+      ...(user.userId && { userId: user.userId }),
+      ...(user.userUuid && { userUuid: user.userUuid }),
+      ...(user.authSource === 'nomis' && user.activeCaseLoadId && { activeCaseLoadId: user.activeCaseLoadId }),
+    })
+    return next()
+  }
+}
```
Thomas-Geraghty, line 32: I wonder if this is worth adding as middleware into the azure-telemetry lib, so that teams can quickly adopt it, and we can change this from a centralised place in future?

server/app.ts
```diff
@@ -35,6 +36,7 @@ export default function createApp(services: Services): express.Application {
   app.use(authorisationMiddleware())
   app.use(setUpCsrf())
   app.use(setUpCurrentUser())
+  app.use(addUserMetadataToTelemetry()) // For services with prison users, add this after middleware that retrieves user caseload data
```
Thomas-Geraghty, line 39: I think this comment probably needs to be on the function itself rather than appended here

brightonsbox: I thought it would be important to see the comment in place when doing a cherry-pick otherwise the ordering issue could easily get missed

server/utils/azureAppInsightsUserMetadata.ts
brightonsbox: It perhaps could be as a follow up step, but given that there are some other changes to how the user token is parsed and processed, it felt important to start here in the template.

andrewrlee (dismissed):
LGTM

server/utils/azureAppInsightsUserMetadata.ts
Thomas-Geraghty: And done - https://github.com/ministryofjustice/hmpps-template-typescript/pull/822/commits/68bd0e0feba584d625b6e808632da8b4a9ca9a2b !

server/app.ts
Thomas-Geraghty: Bumped the comment onto the line above and made it clear it applies to the `addUserMetadataToTelemetry` middleware

andrewrlee (approved):
LGTM

